PR #3, opened 2023-10-10
Fix MeCabTokenizer test

liwii:
Changes from @Alnusjaponica

`"MeCab" in sys.modules` couldn't detect the existence of the module properly, so `pytest.skipif` didn't work. This PR solves the problem.


kennysong:
Tested locally and LGTM!

kennysong (approved):

PR #12, opened 2023-10-10
Allow metrics to take a single string as a parameter

### LINKED ISSUE #5: Rename langcheck.eval to langcheck.metrics
`eval` is a built-in global function in Python.

### LINKED ISSUE #6: Make all metrics accept a single string value as well
Just a bit of syntactic sugar. Right now you need to provide a list of strings.

kennysong:
Resolves #6 .

This is mostly syntactic sugar, and also closes a TODO in the codebase to refactor the validation logic for all metrics.


tests/eval/test_text_structure.py
```diff
@@ -189,13 +227,22 @@ def test_contains_all_strings(generated_outputs, strings, case_sensitive,
                                       case_sensitive)
     assert eval_value.metric_name == 'contains_all_strings'
     assert eval_value.prompts is None
-    assert eval_value.generated_outputs == generated_outputs
+    if isinstance(generated_outputs, str):
```
liwii, line 230: Would it be useful to make a util function for this pattern because it appears so frequently?

src/langcheck/eval/_validation.py
```diff
@@ -0,0 +1,145 @@
+from __future__ import annotations
+
+from typing import List, Optional
+
+
+def validate_parameters_reference_based(
+        generated_outputs: List[str] | str,
```
liwii, line 7: What do you think about add name to `List[str] | str` and use that as the type annotation in each function?
I think it would be easier to read, and also make it easier to add more syntax sugar in the future.

liwii (commented):
Code LGTM, just left a few comments about repeated patterns!!

liwii (commented):
Code LGTM, just left a few comments about repeated patterns!!

src/langcheck/eval/_validation.py
kennysong: `List[str] | str` isn't too long, and I think it's more clear for users reading the documentation who aren't familiar with custom types in our codebase.

tests/eval/test_text_structure.py
kennysong: Sure, refactoring into a function in `tests/utils.py`. Makes the tests a bit more readable in this file.

src/langcheck/eval/_validation.py
liwii: Sounds reasonable, let's leave it as is then.

liwii (approved):
LGTM!!

PR #14, opened 2023-10-11
Pin the PyTorch version in the dev container

liwii:
Ref: https://github.com/pytorch/pytorch/issues/110819

PyTorch 2.1.0 could cause segmentation fault in aarch64 devices. We pin the PyTorch version to 2.0.1 until the problem is fixed.


kennysong (approved):
LGTM!

PR #20, opened 2023-10-11
Rename semantic_sim to semantic_similarity

kennysong:
`semantic_similarity` is more clear (I'm not sure why I named it `semantic_sim` originally). 

This is also more consistent with our `ai_disclaimer_similarity()` metric.


liwii (approved):
LGTM!!

PR #22, opened 2023-10-13
fix contributing guide for zsh

shibuiwilliam:
# What
- Fix dev installation for zsh, probably used by many Mac users.

# Why
- You need to escape bracket for zsh.


liwii (approved):
LGTM, thanks for your contribution!!!

PR #25, opened 2023-10-17
Keyboard typo

Alnusjaponica:
## Motivation

implement `langcheck.augment.keyboard_typo()`


src/langcheck/augment/en/_keyboard_typo.py
```diff
@@ -0,0 +1,22 @@
+from typing import List
+
+import nlpaug.augmenter.char as nac
+
+
+def keyboard_typo(
+    texts: List[str],
```
liwii, line 7: Could you also make it accept single `str`? (Arguments for metrics are so updated recently).

src/langcheck/augment/en/_keyboard_typo.py
```diff
@@ -0,0 +1,22 @@
...
+    **kwargs,
+) -> List[str]:
+    '''Generate keyboard typo perturbed texts for augmentation.
+
+    Args:
+        texts: List of texts to be augmented.
```
liwii, line 13: I think showing some arguments of `nlpaug.augmenter.char.keyboard.KeyboardAug` explicitly here is helpful for langcheck users. At least

- aug_char_p (float)
- aug_char_min (int)
- aug_char_max (int)
- aug_word_p (float)
- aug_word_min (int)
- aug_word_max (int)

seem general enough to add to the list of langcheck arguments. WDYT?

src/langcheck/augment/en/_keyboard_typo.py
```diff
@@ -0,0 +1,22 @@
...
+        **kwargs: Arbitrary keyword arguments, which will be passed to
+            :class:`nlpaug.augmenter.char.keyboard.KeyboardAug`. See the
+            documentation for more details.
```
liwii, line 16: Could you add the link to their readthedocs?
https://nlpaug.readthedocs.io/en/latest/augmenter/char/keyboard.html

src/langcheck/augment/en/_keyboard_typo.py
```diff
@@ -0,0 +1,22 @@
...
+
+    Returns:
+        A list of perturbed texts.
+    '''
+    aug = nac.KeyboardAug()
+    return aug.augment(texts, **kwargs)
```
liwii, line 22: Seemingly these `kwargs` should be passed to `nac.KeyboardAug()` constructor?

tests/augment/test_keboard_typo.py
```diff
@@ -0,0 +1,20 @@
...
+
+import pytest
+
+from langcheck.augment import keyboard_typo
+
+
+@pytest.mark.parametrize(
+    "texts, expected",
+    [
+        (["Hello, world!"], ["HF;lo, eorlE!"]),
```
liwii, line 12: The perturbation seems too much as the default setting to me. How about using some values like `aug_char_p=0.1` as the default?

```python
>>> import nlpaug.augmenter.char as nac
>>> import random
>>> random.seed(42)
>>> nac.KeyboardAug().augment(['Hello, world!!'])
['HF;lo, eorlE!!']
>>> nac.KeyboardAug(aug_char_p=0.1).augment(['Hello, world!!'])
['He?lo, w*rld!!']
```

liwii: `include_special_char=False, include_numeric=False` should be reasonable as the default setting too

liwii (changes requested):
Thanks for quick implementation! Left a few comments.

Alnusjaponica:
@liwii I have resolved all the comments. PTAL.

src/langcheck/augment/__init__.py
```diff
@@ -0,0 +1,5 @@
+from langcheck.augment import en
+
+__all__ = [
+    "en",
```
liwii, line 4: From `langcheck.metrics` you can also directly call English metrics without `en` module. Could you also do that in `augment` for consistency? (I do not think it is a very good design, but let's re-design them altogether later)

src/langcheck/augment/en/_keyboard_typo.py
```diff
@@ -0,0 +1,47 @@
+from __future__ import annotations
+
+from nlpaug.augmenter.char.keyboard import KeyboardAug
+
+
+def keyboard_typo(
+    texts: list[str] | str,
```
liwii, line 7:
```suggestion
    texts: List[str] | str,
```
For consistency with other parts of the codebase?

src/langcheck/augment/en/_keyboard_typo.py
```diff
@@ -0,0 +1,47 @@
...
+    **kwargs,
+) -> list[str]:
```
liwii, line 9:
```suggestion
) -> List[str]:
```

src/langcheck/augment/en/_keyboard_typo.py
```diff
@@ -0,0 +1,47 @@
...
+        listed below:
+            - ``aug_char_p`` (float): Percentage of character (per token)
+              will be augmented.
+            - ``aug_char_min`` (int): Minimum number of character will be
+              augmented.
+            - ``aug_char_max`` (int): Maximum number of character will be
+              augmented.
+            - ``aug_word_p`` (float): Percentage of word will be augmented.
+            - ``aug_word_min`` (int): Minimum number of word will be augmented.
+            - ``aug_word_max`` (int): Maximum number of word will be augmented.
```
liwii, line 29:
```suggestion

        - ``aug_char_p`` (float): Percentage of character (per token)
          will be augmented.
        - ``aug_char_min`` (int): Minimum number of character will be
          augmented.
        - ``aug_char_max`` (int): Maximum number of character will be
          augmented.
        - ``aug_word_p`` (float): Percentage of word will be augmented.
        - ``aug_word_min`` (int): Minimum number of word will be augmented.
        - ``aug_word_max`` (int): Maximum number of word will be augmented.

```

There are some errors when building the doc. Seemingly unindenting the bullet list & inserting blanks before & after that will resolve the issue.

liwii (approved):
LGTM after resolving the last few points!!

src/langcheck/augment/en/_keyboard_typo.py
Alnusjaponica: Using `typing.List` is deprecated by [PEP585](https://peps.python.org/pep-0585/#:~:text=Importing%20those%20from%20typing%20is%20deprecated.). I'll create an issue to handle them but should I restore these change for now?

liwii: Ah good to here, then let's leave this line unchanged. Thanks for creating the issue for that!

src/langcheck/augment/__init__.py
Alnusjaponica: Import path updated

src/langcheck/augment/en/_keyboard_typo.py
Alnusjaponica: I fixed the sphinx error.

liwii:
Did a final review and LGTM! Feel free to merge, thanks for the changes!!

PR #26, opened 2023-10-18
Update the translation model

liwii:
Updated the translation model used for Japanese factual consistency to the one with apache-2.0 license.

Ref: https://huggingface.co/Helsinki-NLP/opus-mt-ja-en/tree/main


kennysong (approved):
LGTM!

PR #27, opened 2023-10-19
Tutorial based on an email generator

yosukehigashi:
Tutorial for how to use LangCheck to evaluate a simple email generator. The sections are:
1. Building the email generator with the OpenAI API, and seeing some example generations
2. Using the reference free metrics to evaluate the generated emails
3. Using reference based metrics


kennysong:
Btw, you need to add the new .md page into the documentation tree so Sphinx finds it when building the docs. 

I pushed a commit to do this (lmk if you have any questions): https://github.com/citadel-ai/langcheck/pull/27/commits/369500c5137ab9366f05f06ff8935127d3bde214

To see the docs locally you can follow step 2 & 3 here: https://langcheck.readthedocs.io/en/latest/contributing.html#documentation

docs/index.md
```diff
@@ -10,6 +10,7 @@ Get started with the docs below.
 installation.md
 quickstart.md
 metrics.md
+tutorials.md
```
kennysong, line 13: (index.md is basically the root of the documentation tree, so Sphinx will start recursively collecting pages linked from here)

docs/tutorial_email_generator.md
```diff
@@ -0,0 +1,225 @@
...
+recipient_name = "Google"
+content_description = """
+- I want to apply for the role
+- I have many years of experience in AI, making me a strong candidate
+- Ready to start ASAP
+"""
+email = generate_email(subject, recipient_name, content_description)
+print(email)
+```
+```text
```
kennysong, line 96: I think code blocks don't do text wrapping, which makes this a bit hard to read.

Maybe we can use a quote block for this or something?

https://myst-parser.readthedocs.io/en/latest/syntax/typography.html#quotations

docs/tutorial_email_generator.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+## Evaluating with LangCheck (No References)
+
+At this point in our development process of the email generator, we would start iterating on the prompt template to try and improve the outputs. This is where LangCheck can come in and help - rather than having to manually prompt the app and check the outputs each time, we can automatically evaluate the outputs with the various LangCheck metrics!
+
+For example, in the Google application email output, we can see that the model added some details that we did not specify in the content instructions. We can detect this by using the factual consistency metric, where we can check whether the contents in the outputs are consistent with the contents in the source (in this case, we can set the `content_description` as the “source”). Indeed, we see that this output has a low factual consistency score.
+
+```python
+factual_consistency = langcheck.metrics.factual_consistency(generated_outputs=[email], sources=[content_description])
```
kennysong, line 122: You can provide a single string to metrics now! No need to wrap it in a list

docs/tutorial_email_generator.md
```diff
@@ -0,0 +1,225 @@
...
+```python
+generated_emails = [item["generated_email"] for item in [japan_trip, running_late, google_application]]
+reference_emails = [item["reference"] for item in [japan_trip, running_late, google_application]]
+
+rougeL = langcheck.metrics.rougeL(generated_outputs=generated_emails, reference_outputs=reference_emails)
+semantic_similarity = langcheck.metrics.semantic_similarity(generated_outputs=generated_emails, reference_outputs=reference_emails)
+
+langcheck.plot.scatter(semantic_similarity, rougeL)
+```
+![MetricValue output](_static/email_tutorial_scatter.png)
```
kennysong, line 223: Maybe a screenshot with a tooltip would be better?

docs/tutorial_email_generator.md
```diff
@@ -0,0 +1,225 @@
...
+####
+
+factual consistency = 0.55
+```
+
+## Evaluating with LangCheck's Reference-Based Metrics
+
+The metrics that we have examined so far have been text structure and reference free metrics, meaning that we don’t need a reference output to evaluate the email generator’s outputs. However, if we do have some reference outputs, we would be able to leverage the reference based metrics as well.
+
+For all of the email prompts, I wrote out the emails that I personally would’ve written (see here). Let’s now assume that my email writing skills are good enough to call these the reference outputs, and compute some of LangCheck’s reference based metrics. Below, we compute the `rougeL` and `semantic_similarity` metrics, and show them both in a scatter plot.
```
kennysong, line 212: "see here" isn't linked to anywhere

docs/tutorial_email_generator.md
```diff
@@ -0,0 +1,225 @@
...
+reference_emails = [item["reference"] for item in [japan_trip, running_late, google_application]]
+
+rougeL = langcheck.metrics.rougeL(generated_outputs=generated_emails, reference_outputs=reference_emails)
+semantic_similarity = langcheck.metrics.semantic_similarity(generated_outputs=generated_emails, reference_outputs=reference_emails)
+
+langcheck.plot.scatter(semantic_similarity, rougeL)
+```
+![MetricValue output](_static/email_tutorial_scatter.png)
+
+Due to the longform nature of these generated emails, we can see that the `rougeL` values tend to be a bit low, even though the `semantic_similarity` values are high.
```
kennysong, line 225: It might be cool to do something with automatically generating the email subject too.

docs/tutorial_email_generator.md
```diff
@@ -0,0 +1,225 @@
+# Evaluating an Email Generator
```
kennysong, line 1: We can do this separately, but it would be pretty cool to have a UI like Koki's Q&A app that shows a bunch of the langcheck metrics next to the generated email

kennysong: Could also add a link to a Colab notebook so people can try the tutorial easily, but I think having a UI would be even better

docs/tutorial_email_generator.md
```diff
@@ -0,0 +1,225 @@
...
+Thank you for considering my application. I look forward to the possibility of joining Google and contributing to its mission of pushing the boundaries of technology.
+
+Sincerely,
+[Your Name]
+```
+
+## Evaluating with LangCheck (No References)
+
+At this point in our development process of the email generator, we would start iterating on the prompt template to try and improve the outputs. This is where LangCheck can come in and help - rather than having to manually prompt the app and check the outputs each time, we can automatically evaluate the outputs with the various LangCheck metrics!
+
```
kennysong, line 118: Edited to be more concise

```suggestion
At this point in our development process, we would start iterating on the prompt template to try and improve the outputs. This is where LangCheck can come in and help - rather than having to manually prompt the app and check the outputs each time, we can automatically evaluate the outputs with the various LangCheck metrics!
```

docs/tutorial_email_generator.md
```diff
@@ -0,0 +1,225 @@
...
+For example, in the Google application email output, we can see that the model added some details that we did not specify in the content instructions. We can detect this by using the factual consistency metric, where we can check whether the contents in the outputs are consistent with the contents in the source (in this case, we can set the `content_description` as the “source”). Indeed, we see that this output has a low factual consistency score.
+
```
kennysong, line 120: Edited to be more concise

```suggestion
For example, in the Google application email output, we saw that the model added some details that we did not specify in the instructions. We can detect this by using the factual consistency metric, which checks whether the output text is consistent with the source text (in this case, we can set the `content_description` as the “source”). Indeed, we see that this output has a low factual consistency score.
```

docs/tutorial_email_generator.md
```diff
@@ -0,0 +1,225 @@
...
+factual_consistency = langcheck.metrics.factual_consistency(generated_outputs=[email], sources=[content_description])
+print(f"factual consistency = {factual_consistency.metric_values[0]:.2f}")
+```
+
+```text
+factual consistency = 0.37
+```
+
+There are many other factors other than factual consistency that we would care about in our email generator app. As examples, we want the email to also be:
+- Not offensive/toxic
```
kennysong, line 131: Edited to be more concise

```suggestion
There are many factors other than factual consistency that we might care about in our email generator app. For example, we also want the email to be:
```

docs/tutorial_email_generator.md
```diff
@@ -0,0 +1,225 @@
...
+[Your Name]
+####
+
+factual consistency = 0.55
+```
+
+## Evaluating with LangCheck's Reference-Based Metrics
+
+The metrics that we have examined so far have been text structure and reference free metrics, meaning that we don’t need a reference output to evaluate the email generator’s outputs. However, if we do have some reference outputs, we would be able to leverage the reference based metrics as well.
+
```
kennysong, line 211: Edited to be more concise

```suggestion
The metrics that we have examined so far have been reference-free and source-based metrics, meaning that we don’t need a reference output to evaluate the email generator’s outputs. However, if we do have some reference outputs, we can leverage the reference-based metrics as well.
```

kennysong (changes requested):
Pretty nice tutorial! Left a few comments

docs/tutorial_email_generator.md
yosukehigashi: The scatter plot doesn't look that great with long texts IMO - WDYT?
image

docs/tutorial_email_generator.md
yosukehigashi: hmm I actually kind of liked how it doesn't do text wrapping, since then it takes up much less screen real estate. I also didn't think it was that important for the email to be easy to read, since the exact contents of the emails aren't really that important 😅 Here's how it looks with quote blocks though vs. the code text block from before. WDYT?

image

image

docs/tutorial_email_generator.md
yosukehigashi: Oh nice! Removed the list wrapper

docs/tutorial_email_generator.md
yosukehigashi: oh oops, I'll link this to a colab!

docs/tutorial_email_generator.md
yosukehigashi: I think a colab would be nice so that users can change the prompts and reference outputs easily. Having a UI in addition to that would be cool too, but I'll punt on that for now!

docs/tutorial_email_generator.md
kennysong: I think a first time reader will want to identify the hallucinations in the email that you pointed out, which was kind of hard with scrolling. It's pretty easy to visually skip the block if someone's not interested in reading. 

(We could also edit the email a bit so it's not as long)

docs/tutorial_email_generator.md
kennysong: Oooh yeah, we should definitely fix that in general. Can you try something like this? I'm on plane wifi right now so it's hard to experiment. 

https://github.com/plotly/plotly.js/issues/1964#issuecomment-612641307

You can add it here:

https://github.com/citadel-ai/langcheck/blob/369500c5137ab9366f05f06ff8935127d3bde214/src/langcheck/plot/_scatter.py#L48

We can wrap to 3 lines and elide the rest or something.

docs/tutorial_email_generator.md
yosukehigashi: Got it - changed everything to use quote blocks

docs/tutorial_email_generator.md
yosukehigashi: linked to the colab

docs/tutorial_email_generator.md
yosukehigashi: I made it be a max of 5 lines and `...` if there are more, e.g.
Screen Shot 2023-10-23 at 18 10 20

docs/tutorial_email_generator.md
yosukehigashi: I changed the subject to be auto-generated too by making it follow the format
```
        ###
        Subject: <appropriate subject>

        <Email content>
        ###
```

docs/tutorial_email_generator.md
yosukehigashi: added a colab and linked it at the top of the tutorial!

docs/tutorial_email_generator.md
kennysong: Thanks!

docs/tutorial_email_generator.md
kennysong: Nice!!

docs/tutorial_email_generator.md
```diff
@@ -0,0 +1,266 @@
...
+There are many factors other than factual consistency that we might care about in our email generator app. For example, we also want the email to be:
+- Not offensive/toxic
+- Fluent
+- Generally have a positive sentiment (though maybe not always)
+- Not contain any AI disclaimer language (i.e. language similar to “As an AI model, I don’t have opinions”)
+
+Here is how we can use LangCheck to check all of these aspects.
+```python
+emails = [item["generated_email"] for item in [japan_trip, running_late, google_application]]
+content_descriptions = [item["content_description"] for item in [japan_trip, running_late, google_application]]
```
kennysong, line 142: Nit: Can we format the code blocks with yapf so they're not super long?

Also, this is a style thing, but I feel like we don't need to specify the parameter names in our metric functions since they get super long. At least for the reference-free metrics with a single argument.

docs/tutorial_email_generator.md
kennysong: Nice!

docs/tutorial_email_generator.md
```diff
@@ -220,6 +217,50 @@ semantic_similarity = langcheck.metrics.semantic_similarity(generated_outputs=ge
...
+    for item in [japan_trip, running_late, google_application]
+]
+reference_emails = [
+    item["reference"]
+    for item in [japan_trip, running_late, google_application]
+]
+
+# Check that the generated email contains the string "Subject: "
+contains_subject = langcheck.metrics.contains_any_strings(
+    generated_outputs=generated_emails, strings=["Subject: "])
```
kennysong, line 240: We could also use `contains_regex()` with your regex below. But I think `contains_any_strings()` is fine too

kennysong (approved):
LGTM after resolving comments!

docs/tutorial_email_generator.md
yosukehigashi: > Nit: Can we format the code blocks with yapf so they're not super long?

Done!

> Also, this is a style thing, but I feel like we don't need to specify the parameter names in our metric functions since they get super long. At least for the reference-free metrics with a single argument.

Makes sense, I removed the param names for the metrics with just one argument. For the ones with two arguments, I kept the names since we have a mix of `sources=` and `reference_outputs=` in this tutorial, so I think it's helpful to make the differences clear

docs/tutorial_email_generator.md
yosukehigashi: Changed to `contains_regex()`!

docs/tutorial_email_generator.md
kennysong: LGTM!

PR #33, opened 2023-10-25
Handle cases where an evaluator fails for certain inputs

yosukehigashi:
There are certain cases where an evaluator may fail for a certain reason (e.g. OpenAI failing to return a response). Rather than failing completely, we will now still return the scores for the inputs that can be evaluated.

The `metric_values` list in the `MetricValue` object is now allowed to contain `None` values. The majority of the changes in this PR are to account for that change.


src/langcheck/metrics/metric_value.py
```diff
@@ -162,9 +163,13 @@ def __post_init__(self):
...
             raise ValueError(f'Invalid threshold operator: {self.threshold_op}')
 
+        if self.threshold is None:
+            raise ValueError("A threshold of `None` is not supported.")
+
+        # Set the result to `False` if the metric value is `None`
         self._threshold_results = [
             operators[self.threshold_op](x, self.threshold)
-            for x in self.metric_values
+            if x is not None else False for x in self.metric_values
```
liwii, line 172: Maybe at least for `!==` the value shouldn't be `False`...?

src/langcheck/metrics/en/_openai.py
```diff
@@ -45,10 +45,7 @@ def get_score(self, prompt: str) -> float:
 
         Returns:
             Score associated with the given prompt based the resulting
-                assessment
-
-        Raises:
-            AssertionError: If the OpenAI API returns an unrecognized assessment
+                assessment, `None` if the score could not be computed
```
liwii, line 48: Would it be helpful to explicitly say that when the score is not computed there are some issues about OpenAI APIs?

liwii (commented):
Left a few comments!!

src/langcheck/metrics/metric_value.py
yosukehigashi: hmm but the way that we frame it is that `True` means that it's a "pass", and I don't think `None` should ever be a "pass"? E.g. if the user writes `toxicity != 1`, I think a `None` value should fail that check

(Maybe we should output a warning any time a user calls a comparison operator on a MetricValue with Nones?)

src/langcheck/metrics/en/_openai.py
yosukehigashi: It could be due to OpenAI not returning a value in the format specified in the prompt as well, which is a little different from an issue with the API. Added some more detail in the function's docstring

src/langcheck/metrics/metric_value.py
liwii: > output a warning any time a user calls a comparison operator on a MetricValue with Nones

That sounds good to me, could you work on that?

yosukehigashi: Added the warning!

liwii (approved):
LGTM!!

PR #35, opened 2023-10-27
Chain of thought prompt for OpenAI evaluators

yosukehigashi:
This updates all of the OpenAI-based evaluators to (1) first output an "unstructured assessment" using a normal API call, and then to (2) output the "structured assessment" based on the unstructured assessment using the function calling API. 

Allowing the model to first reason about the problem, and then come up with the final answer, results in much better performance. I used the "Take a deep breath and work on this problem step-by-step." prompt that DeepMind found to be an effective CoT prompt ([paper](https://arxiv.org/pdf/2309.03409.pdf)). Improvements to the factual consistency benchmark are
| Data | OpenAI New Prompt | OpenAI Old Prompt |
|-------------|---------|---------|
| **QAGS-XSUM**    | Pearson correlation = 0.40<br>Spearman correlation = 0.40<br>Kendall-Tau correlation = 0.38 | Pearson correlation = 0.31<br>Spearman correlation = 0.32<br>Kendall-Tau correlation = 0.31  |
| **QAGS-CNN**    | Pearson correlation = 0.67<br>Spearman correlation = 0.59<br>Kendall-Tau correlation = 0.52 | Pearson correlation = 0.42<br>Spearman correlation = 0.37<br>Kendall-Tau correlation = 0.32  |

This also has the nice benefit of outputting explanations for each score, which I added as a field to the `MetricValue` object., e.g.
image


kennysong:
Apparently the "take the deep breath" prompts is optimized for Palm. Should we try the other prompts here?

image

kennysong:
Oh I guess those models are used to generate the prompt, but all of the prompts were evaluated using either PaLM or Bison.

yosukehigashi:
> Oh I guess those models are used to generate the prompt, but all of the prompts were evaluated using either PaLM or Bison.

Yeah this was mostly on Google's models, but I've anecdotally heard that the "Take a deep breath" trick works for OpenAI as well (e.g. some people are adding it to their Custom Instructions in ChatGPT). This optimization process was also run on an arithmetic dataset so the other prompts are pretty math-y, and the "Take a deep breath" one seems the most general.

Longer term, once we have some robust datasets, we can try to do some prompt optimization ourselves for each use case.

src/langcheck/metrics/metric_value.py
```diff
@@ -22,6 +22,7 @@ class MetricValue(Generic[NumericType]):
     generated_outputs: List[str]
     reference_outputs: Optional[List[str]]
     sources: Optional[List[str]]
+    explanations: Optional[List[Optional[str]]]
```
kennysong, line 25: Why is it `Optional[str]` instead of `str`?

src/langcheck/metrics/en/reference_free_text_quality.py
```diff
@@ -445,7 +496,23 @@ def _prompt(gen_output: str) -> str:
...
+        '''
+
+    def _function_call_prompt(long_assessment: str) -> str:
+        return f'''
+        The following is an assessment on the toxicity of a statement:
+        ************
+        [Assessment]: {long_assessment}
+        ************
+
+        Save the resulting assessment. The available assessments are:
```
kennysong, line 510: What does "Save the resulting assessment" mean?

src/langcheck/metrics/metric_value.py
kennysong: Ah I see. How about adding a comment since it's different from the other types?

```suggestion
        explanations: Optional[List[Optional[
            str]]]  # An explanation can be None if the metric could not be computed
```

kennysong (approved):
LGTM after resolving comments!

kennysong (commented):
LGTM after resolving comments!

src/langcheck/metrics/en/reference_free_text_quality.py
yosukehigashi: The way that we get a structured response is by forcing OpenAI to call a made-up function with an argument that has a structured schema. E.g. for sentiment, we force OpenAI to call `save_sentiment_assessment(sentiment_assessment)`, where `sentiment_assessment` has to be one of `['Positive', 'Neutral', 'Negative']`.

So I included 'Save the resulting assessment' in the prompt so that it's consistent with how we're using the function calling API

kennysong: Got it!

PR #23, opened 2023-10-17
Clean up noqa in docstrings

Alnusjaponica:
## Motivation
Resolve #21  and Reformat NOQA comment as in [flake8 official docs](https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors)

PR #37, opened 2023-10-31
Fix docs warning

Alnusjaponica:
## Motivation
Follow-up https://github.com/citadel-ai/langcheck/pull/23

Currently, we have following error when building docs:
```
...
docstring of langcheck.metrics.en.reference_free_text_quality.fluency:7: ERROR: Unknown target name: "this example https://langcheck.readthedocs.io/en/latest/metrics.html#computing-metrics-with-openai-models>".
...
```
and this PR fixes them.


liwii (approved):
Confirmed that these errors disappear but the links are still accessible, LGTM!!

Issue #38, opened 2023-11-03
Visualize thresholds when plotting MetricValueWithThresholds


kennysong:
When plotting a MetricValueWithThreshold value like `(toxicity_values > 0.5).scatter()`, we should draw a vertical line in the chart where the threshold is.

We could do this for the double scatter plot (`langcheck.plot.scatter(toxicity_values > 0.5, sentiment_values > 0.5)`) and the histogram as well (`(toxicity_values > 0.5).histogram()`).

Vela-zz:
@kennysong  Hi Song, I'm trying to add code for this feature now, but before I  complete it, I want to ask is there any plan to support more Filter Syntax in the future？

```
toxicity_values > 0.5 and toxicity_values < 0.7
#
toxicity_values < 0.3 | toxicity_values > 0.7

```
I think it is also a new feature that can be added.

kennysong:
Hi @Vela-zz, thanks for the PR! I don't think we're planning to support those multi-condition thresholds right now, so we can keep your PR focused on the existing binary threshold case.

If you're looking for other issues to work on, happy to chat more [on Discord](https://discord.com/invite/Bkndx9RXqw)!

cc @yosukehigashi

PR #40, opened 2023-11-03
add line when plot MetricValueWithThresholds

Vela-zz:
#38 add dash line for MetricValueWithThresholds

image
 
image

image

### DIFF
--- a/src/langcheck/plot/_histogram.py
+++ b/src/langcheck/plot/_histogram.py
@@ -3,7 +3,7 @@ import math
 import plotly.express as px
 from dash import Dash, Input, Output, dcc, html
 
-from langcheck.metrics.metric_value import MetricValue,MetricValueWithThreshold
+from langcheck.metrics.metric_value import MetricValue, MetricValueWithThreshold
 from langcheck.plot._css import GLOBAL_CSS
 from langcheck.plot._utils import Axis, _plot_threshold
 
@@ -69,7 +69,8 @@ def histogram(metric_value: MetricValue, jupyter_mode: str = 'inline') -> None:
         # Plot the histogram
         fig = px.histogram(df, x=metric_value.metric_name)
         if isinstance(metric_value, MetricValueWithThreshold):
-            _plot_threshold(fig, metric_value.threshold_op, metric_value.threshold, Axis.vertical)
+            _plot_threshold(fig, metric_value.threshold_op,
+                            metric_value.threshold, Axis.vertical)
         # Manually set the number of bins in the histogram. We can't use the
         # nbins parameter of px.histogram() since it's just a suggested number
         # of bins. See: https://community.plotly.com/t/histogram-bin-size-with-plotly-express/38927/5  # NOQA: E501
--- a/src/langcheck/plot/_histogram.py
+++ b/src/langcheck/plot/_histogram.py
@@ -3,7 +3,7 @@ import math
 import plotly.express as px
 from dash import Dash, Input, Output, dcc, html
 
-from langcheck.metrics.metric_value import MetricValue
+from langcheck.metrics.metric_value import MetricValue,MetricValueWithThreshold
 from langcheck.plot._css import GLOBAL_CSS
 from langcheck.plot._utils import Axis, _plot_threshold
 
@@ -68,10 +68,8 @@ def histogram(metric_value: MetricValue, jupyter_mode: str = 'inline') -> None:
     def update_figure(num_bins):
         # Plot the histogram
         fig = px.histogram(df, x=metric_value.metric_name)
-        if 'threshold_test' in df.columns:
-            # Get threshold text from dataframe
-            threshold_text = df.get("threshold_test").unique().tolist()[0]
-            _plot_threshold(fig, threshold_text, Axis.vertical)
+        if isinstance(metric_value, MetricValueWithThreshold):
+            _plot_threshold(fig, metric_value.threshold_op, metric_value.threshold, Axis.vertical)
         # Manually set the number of bins in the histogram. We can't use the
         # nbins parameter of px.histogram() since it's just a suggested number
         # of bins. See: https://community.plotly.com/t/histogram-bin-size-with-plotly-express/38927/5  # NOQA: E501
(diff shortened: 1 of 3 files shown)


yosukehigashi:
Thank you for your contribution!! I'll review this properly tomorrow, but from a quick glance this looks great 😄

src/langcheck/plot/_scatter.py
```diff
@@ -288,6 +292,17 @@ def update_figure(filter_generated_outputs, filter_reference_outputs,
                          x=metric_value.metric_name,
                          y=other_metric_value.metric_name,
                          hover_data=hover_data)
+        # Paint threshold
```
yosukehigashi, line 295:
```suggestion
        # Draw threshold
```
nit: I think "draw" is more natural than "paint" here

src/langcheck/plot/_scatter.py
```diff
@@ -288,6 +292,17 @@ def update_figure(filter_generated_outputs, filter_reference_outputs,
...
+        if 'threshold_test' in df.columns:
+            # Get threshold text from dataframe
+            threshold_text = df.get("threshold_test").unique().tolist()[0]
+            _plot_threshold(fig, threshold_text, Axis.vertical)
+
+        if 'threshold_test' in other_metric_value.to_df().columns:
+            # Paint other metrics threshold if exsist
```
yosukehigashi, line 302:
```suggestion
            # Draw other metric's threshold if it exists
```

src/langcheck/plot/_utils.py
```diff
@@ -0,0 +1,33 @@
...
+from plotly.graph_objects import Figure
+
+
+class Axis(Enum):
+    vertical = 0
+    horizontal = 1
+
+
+def _plot_threshold(fig: Figure, threshold_text: str, direction: Axis):
+    # Analyize threshold condition
```
yosukehigashi, line 13:
```suggestion
    # Analyze threshold condition
```

src/langcheck/plot/_utils.py
```diff
@@ -0,0 +1,33 @@
...
+    pattern = r'(<|<=|>|>=|==)\s([0-9]\.[0-9]+)'
+    match = re.search(pattern, threshold_text)
+    if match:
+        operator, threshold = match.groups()
+    else:
+        raise ValueError('Threshold not Found!')
+    operator, threshold = operator[0], float(threshold)
+    if direction == Axis.horizontal:  # Paint in row
```
yosukehigashi, line 21:
```suggestion
    if direction == Axis.horizontal:  # Draw a horizontal line
```
Is this what you meant? (I'm not quite sure what "in row" meant here)

src/langcheck/plot/_utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class Axis(Enum):
+    vertical = 0
+    horizontal = 1
+
+
+def _plot_threshold(fig: Figure, threshold_text: str, direction: Axis):
+    # Analyize threshold condition
+    pattern = r'(<|<=|>|>=|==)\s([0-9]\.[0-9]+)'
```
yosukehigashi, line 14: This currently fails in cases where the threshold is not a decimal number, e.g. `(flesch_kincaid_grade_values > 5).scatter()`. Instead of extracting the `operator` and `threshold` from the threshold text, how about getting them directly from the `MetricValueWithThreshold` object?

e.g. `_plot_threshold` could be implemented as
```
def _plot_threshold(fig: Figure, operator: str, threshold: Union[float, int], direction: Axis):
    threshold_text = f'{operator} {threshold}'
    if direction == Axis.horizontal:
        ...
```

and then where `_plot_threshold` gets called, it can look like
```
if isinstance(metric_value, MetricValueWithThreshold):
    _plot_threshold(fig, metric_value.threshold_op, metric_value.threshold, Axis.horizontal)
```

yosukehigashi (changes requested):
Thanks for the PR! Please take a look at the comments, and let me know if anything is unclear or you want to chat!

src/langcheck/plot/_utils.py
Vela-zz: Yes, I mean draw a horizontal line here, the old version here is direction == 0 or direction == 1 here, but i feel ambiguity here. Then i create the Axis class to do this.forget to rewrite the comment here. your comment applied

Vela-zz:
@yosukehigashi  adjust how _plot_threshold get value from metric_value, and add extral code to histograpm() plot to prevent this happens
image

yosukehigashi:
I'm not sure if you're able to merge, so I'll go ahead and merge this. Thanks for the contribution @Vela-zz!

src/langcheck/plot/_utils.py
yosukehigashi: Thanks for the explanation! Agreed that writing `Axis.horizontal` is much clearer than just `1`

src/langcheck/plot/_utils.py
yosukehigashi: Thanks for the fix!

src/langcheck/plot/_histogram.py
```diff
@@ -67,15 +68,22 @@ def histogram(metric_value: MetricValue, jupyter_mode: str = 'inline') -> None:
...
         # Manually set the number of bins in the histogram. We can't use the
         # nbins parameter of px.histogram() since it's just a suggested number
         # of bins. See: https://community.plotly.com/t/histogram-bin-size-with-plotly-express/38927/5  # NOQA: E501
         start = math.floor(df[metric_value.metric_name].min())
         end = math.ceil(df[metric_value.metric_name].max())
         step_size = (end - start) / int(num_bins)
         fig.update_traces(xbins={'start': start, 'end': end, 'size': step_size})
-
+        # Explicitly set the default axis ranges (with a little padding) so that
+        # the range ploted would not influenced by threshold settings
```
yosukehigashi, line 82:
```suggestion
        # the range plotted would not be influenced by threshold settings
```

src/langcheck/plot/_utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class Axis(Enum):
+    vertical = 0
+    horizontal = 1
+
+
+def _plot_threshold(fig: Figure, threshold_op: str,
+                    threshold: Union[float, int], direction: Axis):
+    '''Draw dash line on target figure by giving threshold and axis
```
yosukehigashi, line 15:
```suggestion
    '''Draw a dashed line on the target figure at the specified threshold value
    along either the horizontal or vertical axis.
    
    Args:
        fig: Plotly figure to draw the line on
        threshold_op: A string representing the threshold operator, e.g. '<'
        threshold: Threshold value
        direction: Direction (horizontal or vertical) that the dashed line
            should be drawn
```

yosukehigashi (changes requested):
Thanks for the changes! The code LGTM - just a couple final suggestions on the some of the wording in the comments.

yosukehigashi:
> @yosukehigashi adjust how _plot_threshold get value from metric_value, and add extral code to histograpm() plot to prevent this happens image

Nice catch!

yosukehigashi (approved):

PR #43, opened 2023-11-08
Implement `augment.gender`

Alnusjaponica:
## Motivation 
Implement `augment.gender`


src/langcheck/augment/en/_gender/_gender_pronouns.py
```diff
@@ -0,0 +1,137 @@
...
+    object='them',
+    independent_possessive='theirs',
+    reflexive='themselves',
+)
+
+# Neutral pronouns list were taken from some of the most popular neutral
+# pronouns from the following sources:
+# https://ofm.wa.gov/sites/default/files/public/shr/Diversity/BRGs/Pronouns%20Compilation%20Best%20Practices.pdf
+# https://uwm.edu/lgbtrc/support/gender-pronouns/
+_NEO_PRONOUNS_LIST = [
```
liwii, line 38: It is surprising that there are so many gender-neutral pronouns!!

src/langcheck/augment/en/_gender/_gender.py
```diff
@@ -0,0 +1,122 @@
...
+        return "independent_possessive"
+    return _PRONOUNS_FORM_DICT.get(word.lower())
+
+
+def _replace_pronoun(word: str, tag: str, target_pronouns: _BaseGenderPronouns):
+    # When the word is not a pronoun, return the word itself.
+    if (pronoun_form := _get_pronoun_form(word, tag)) is None:
+        return word
+    # Replace the pronoun with the target pronoun with the same form.
+    replaced_pronoun = getattr(target_pronouns, pronoun_form)
```
liwii, line 55: It might be just a matter of preference, but it would be easier to read for me if `target_pronouns` is a dict and you just get a value by the key-value access. Is there any benefit (e.g. related to types) by using `getattr` here?

src/langcheck/augment/en/_gender/_gender.py
```diff
@@ -0,0 +1,122 @@
...
+    texts: Iterable[str] | str,
+    *,
+    to_gender: str | None = None,
+) -> list[str]:
+    """Replace pronouns with that of specified gender.
+
+        Args:
+            texts: Iterable of texts to be augmented.
+            to_gender: Replacing pronoun type string ('male', 'female' or
+                'neutral'). Default to None and use third-person plural when not
```
kennysong, line 97: One drive-by comment: Instead of `None` representing third-person plural, I think we should just set `to_gender = 'plural'` by default. That makes it a bit more explicit. 

(Also wow, I've never heard of these neutral neopronouns before.)

src/langcheck/augment/en/_gender/_gender_pronouns.py
```diff
@@ -0,0 +1,137 @@
...
+        subject='ae',
+        dependent_possessive='aer',
+        object='aer',
+        independent_possessive='aers',
+        reflexive='aerself',
+    )
+]
+
+_PRONOUNS_DICT = MappingProxyType({
+    None: [_PLURAL_PRONOUNS],
```
liwii, line 133: From @kennysong : Could be better to name it "plural".

liwii (changes requested):
Looks nice!!

Made a few comments, will take another look after the pyright error is resolved.

src/langcheck/augment/en/_gender/_gender_pronouns.py
Alnusjaponica: Updated.

src/langcheck/augment/en/_gender/_gender_pronouns.py
Alnusjaponica: So do I, and there seems more not-well-known ones.

src/langcheck/augment/en/_gender/_gender.py
Alnusjaponica: Thanks for the comment. I updated the default argument.

src/langcheck/augment/en/_gender/_gender.py
Alnusjaponica: I just wanted to make _BaseGenderPronouns immutable, since it should not be change.
Since it is not allowed to inherit `MappingProxyType`, there are no convenient immutable and customizable dict subclass in standard library, as far as I know.

src/langcheck/augment/en/_gender/_gender.py
kennysong: Thanks!

src/langcheck/augment/en/_gender/_gender.py
```diff
@@ -0,0 +1,122 @@
...
+from nltk import pos_tag, word_tokenize
+from nltk.tokenize.treebank import TreebankWordDetokenizer
+
+from langcheck.augment.en._gender._gender_pronouns import (_PRONOUNS_DICT,
+                                                           _BaseGenderPronouns)
+
+try:
+    nltk.data.find('averaged_perceptron_tagger')
+except LookupError:
+    nltk.download('averaged_perceptron_tagger')
```
liwii, line 17: When I ran the code in a brand new dev container, I got the following error. Do we need to add `nltk.download('punkt')` too?

```
>>> langcheck.augment.en.gender(sentence)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/workspaces/langcheck/src/langcheck/augment/en/_gender/_gender.py", line 116, in gender
    return [_replace_gender_pronouns(texts, target_gender)]
  File "/workspaces/langcheck/src/langcheck/augment/en/_gender/_gender.py", line 79, in _replace_gender_pronouns
    tagged_words = pos_tag(word_tokenize(text))
  File "/home/vscode/.local/lib/python3.8/site-packages/nltk/tokenize/__init__.py", line 129, in word_tokenize
    sentences = [text] if preserve_line else sent_tokenize(text, language)
  File "/home/vscode/.local/lib/python3.8/site-packages/nltk/tokenize/__init__.py", line 106, in sent_tokenize
    tokenizer = load(f"tokenizers/punkt/{language}.pickle")
  File "/home/vscode/.local/lib/python3.8/site-packages/nltk/data.py", line 750, in load
    opened_resource = _open(resource_url)
  File "/home/vscode/.local/lib/python3.8/site-packages/nltk/data.py", line 876, in _open
    return find(path_, path + [""]).open()
  File "/home/vscode/.local/lib/python3.8/site-packages/nltk/data.py", line 583, in find
    raise LookupError(resource_not_found)
LookupError: 
**********************************************************************
  Resource punkt not found.
  Please use the NLTK Downloader to obtain the resource:

  >>> import nltk
  >>> nltk.download('punkt')
  
  For more information see: https://www.nltk.org/data.html

  Attempted to load tokenizers/punkt/PY3/english.pickle

  Searched in:
    - '/home/vscode/nltk_data'
    - '/usr/local/nltk_data'
    - '/usr/local/share/nltk_data'
    - '/usr/local/lib/nltk_data'
    - '/usr/share/nltk_data'
    - '/usr/local/share/nltk_data'
    - '/usr/lib/nltk_data'
    - '/usr/local/lib/nltk_data'
    - ''
**********************************************************************
```

src/langcheck/augment/en/_gender/_gender.py
liwii: Got it, let's keep using `getattr` then!!

liwii (changes requested):
Thanks for the changes!!

I tried running the code and left one comment about `nltk` download.

I also found a small bug. When there are more than one sentence or the sentence has an extra space in the last, the space is handled weirdly. Could you investigate the reasons?

```
>>> langcheck.augment.en.gender("Actually she was a fish. Noooo")
['Actually they was a fish . Noooo']
>>> langcheck.augment.en.gender("Actually she was a fish. ")
['Actually they was a fish.']
```

src/langcheck/augment/en/_gender/_gender.py
Alnusjaponica: I added the download step. PTAL.

liwii:
Tell me once the small bug I mentioned here is also fixed!!

Alnusjaponica:
I looks like `nltk` library has a bug in `detokenize` method.
The first bug comes from [here](https://github.com/nltk/nltk/blob/d7b428daa90b41edc5adaf92755cab7aec7f5df2/nltk/tokenize/treebank.py#L362)  and the second one from [here](https://github.com/nltk/nltk/blob/d7b428daa90b41edc5adaf92755cab7aec7f5df2/nltk/tokenize/treebank.py#L376). Shall I 
- create issue in `nltk`'s repository,
- implement detokenize logic by my own in this PR,
-  or create issue here and fix them later?

liwii:
Thanks for the investigation!!

In that case, it should be a lot of work to fix the problem and I don't think it's worth it. 

As you mentioned,
- create an issue in nltk's repository,
- and create an issue here in case we want to fix it before nltk's fix?

After these issues are made and the conflicts are resolved, I'll approve this PR!

Alnusjaponica:
@liwii I opened the two issues I remarked.

liwii (approved):
LGTM, thanks!!

PR #45, opened 2023-11-08
Constraint `tokenizers` version for python >=3.11

yosukehigashi:
The `tokenizers` package first officially added support for Python 3.11 in the [0.13.2 release](https://github.com/huggingface/tokenizers/releases/tag/python-v0.13.2). While it's technically possible to build `tokenizers` from source (although this requires users to have a rust compiler), we found that this fails on x86. So, we add the constraint that `tokenizers >= 0.13.2` for Python 3.11 and above.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
     'sentence_transformers >= 2',
     'sentencepiece>=0.1.95',
     'tomli; python_version < "3.11"',
+    'tokenizers >= 0.13.2; python_version >= "3.11"',
     'torch >= 2',
     'transformers >= 4.6',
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
     'sentence_transformers >= 2',
     'sentencepiece>=0.1.95',
     'tomli; python_version < "3.11"',
+    'tokenizers >= 0.13.2; python_version >= "3.11"',
```
kennysong, line 31: Can you add a link to this PR or the tokenizers release notes?

yosukehigashi: done!

kennysong (approved):
LGTM, thanks!

PR #65, opened 2023-12-06
OpenAI (sort of) reproducible outputs

yosukehigashi:
Following https://cookbook.openai.com/examples/deterministic_outputs_with_the_seed_parameter, we now set the `seed` parameter to make the evaluation results and generated rephrases more deterministic.

Unfortunately, setting the seed doesn't actually guarantee anything, and it's more of a best effort to reproduce the same outputs. Still, it's better than nothing!
image


kennysong:
LGTM!

kennysong (approved):

PR #73, opened 2023-12-10
model config manager class

Vela-zz:
Implement a model manager class for centralized type model management.

The model manager can manage transformer models used in different metric for different language. The manager support list_model_in_use, set_model_for_metric, reset, load_from_file, and save_to_disk, by using this, I think langcheck can 

- [x] a centralized evalution model management
- [x] allow user to use the model they perfered, or let them built the eval pipeline with their own build. 
- [ ] evaluation version control, allow user save evaluation result align with the models evaluate them for better evaluation experinment management and repeated.
- [ ] replace the pydoc in same metric in different language that only has a little differernce with the same one.

As it is a prototype, so I base it on zh-support branch to make it do not influence any  other in-progress feature's updating.

It can be used like this,
屏幕截图 2023-12-10 175028


yosukehigashi:
> Implement a model manager class for centralized type model management.
> 
> The model manager can manage transformer models used in different metric for different language. The manager support list_model_in_use, set_model_for_metric, reset, load_from_file, and save_to_disk, by using this, I think langcheck can
> 
> * [x]  a centralized evalution model management
> * [x]  allow user to use the model they perfered, or let them built the eval pipeline with their own build.
> * [ ]  evaluation version control, allow user save evaluation result align with the models evaluate them for better evaluation experinment management and repeated.
> * [ ]  replace the pydoc in same metric in different language that only has a little differernce with the same one.

Thank you for writing this up!! I like the idea of a Model Manager, and agree that the properties you listed are all nice to have. I feel like it's not quite a "Model Manager" yet though, since it mostly just manages the model name/path.

Concretely, I think that it would be helpful to make the Model Manager return an actual model, rather than just the model name. Here’s an example of what I have in mind:
```
# Define various model loaders
def load_sentence_transformer(name):
    return SentenceTransformer(name)

def load_auto_model_for_sequence_classification(name):
    tokenizer = AutoTokenizer.from_pretrained(name)
    model = AutoModelForSequenceClassification.from_pretrained(name)
    return tokenizer, model

# Other loaders
```
```
# Model Manager loads a model based on the config
class ModelManager:
    def __init__(self, config):
        # Read config

    def get_model(self, language, metric):
        # NOTE: will also need to ensure that we don't reload models unnecessarily
        model_config = self.config[language][metric]
        if model_config['type'] == 'sentence_transformer':
            return load_sentence_transformer(model_config['name'])
        # Handle other types

    def validate_config(self, language='all', metric='all'):
        # Check that the model(s) can be loaded for the specified language and metric given the config

    # Other methods like set, reset, etc.
```

A couple additional benefits that I think this could bring
- This makes it easy for us to add validation code. Otherwise, the user will need to actually run the metric to see if their config is valid or not
- This unifies LangCheck code around loading local models

Let me know what you think!

src/langcheck/metrics/_model_management.py
```diff
@@ -0,0 +1,216 @@
...
+            if revision is None:
+                url = f"https://huggingface.co/api/models/{model_name}"
+            else:
+                url = f"https://huggingface.co/api/models/{model_name}/revision/{revision}"  # NOQA:E501
+            response = requests.get(url)
+            return response.status_code == 200
+
+        config = deepcopy(self.config)
+        for lang, lang_setting in config.items():
+            if language == 'all' or lang == language:
```
yosukehigashi, line 145:
```suggestion
            if language != 'all' and lang != language:
                continue
```
so that we can remove one level of nesting

src/langcheck/metrics/_model_management.py
```diff
@@ -0,0 +1,216 @@
...
+                for metric_name, model_setting in lang_setting.items():
+                    if metric == 'all' or metric_name == metric:
```
yosukehigashi, line 147:
```suggestion
                    if metric != 'all' and metric_name != metric:
                        continue
```
(again, to remove nesting)

yosukehigashi (commented):
Thanks for the change, this looks great overall!

I committed a few cleanup changes directly, and made a few minor comments in this initial review. I'll play around with the model manager a bit more and try out the caching / config interface and send another follow-up review

src/langcheck/metrics/config/metric_config.ini
```diff
@@ -0,0 +1,22 @@
...
+    # the 3 models of different sizes provided BAAI are the best on the
+    # embedding task
+    # Ref: https://huggingface.co/BAAI/bge-base-zh-v1.5
+    # Using this model, it is hard to find two sentence where cos_sim < 0.25.
+    model_name = BAAI/bge-base-zh-v1.5
+    revision = f03589c
+    loader = sentence-transformers
+[[sentiment]]
+    model_name = IDEA-CCNL/Erlangshen-Roberta-110M-Sentiment
+    loader = huggingface
```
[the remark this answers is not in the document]
Vela-zz: thanks, I may choose to change this to hf_text-classifcation to align with the function name in loader. if we always use sentence-transformer for embedding (hope the OOM problem solved), and huggingface for classification, I can update this section to "task = embedding" or "task = classificaiton"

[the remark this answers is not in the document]
yosukehigashi: If we update the format of the config file though, it could break users' code so I think we should try to not change it as much as possible? E.g. if I (as a user) have code that's like
```
model_manager.set_model_for_metric('zh', 'sentiment', 'some_custom_model_name', loader='huggingface')
```
But then LangCheck gets updated to no longer recognize the loader `huggingface`, it'll be a breaking change for me.

A temporary solution for this might be to first launch the Model Manager _without_ the `set_model_for_metric` method, since that's the only one that could cause issues with compatibility. Then, once we have a stable format for the config that we don't expect to change that often, we could add the `set_model_for_metric` to make it customizable for users. What do you think?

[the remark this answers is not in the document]
Vela-zz: yes, change config format frequent make user confused, definitely. Compare with fully remove it, how about make it a private function first, I think we can use set_metric_for_lang in model_manager init phase. The modemanger  set_metric_for_lang one by one when initiated，logically the same.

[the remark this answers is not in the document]
yosukehigashi: Ah good idea, yes let's make it private first then!

src/langcheck/metrics/__init__.py
```diff
@@ -13,6 +14,8 @@
                                               is_json_array, is_json_object,
                                               matches_regex, validation_fn)
 
+_model_manager = ModelManager()
```
yosukehigashi, line 17: Instead of initializing this here, what do you think about creating a separate directory inside of `metrics`? E.g. something like
```
metrics/
    model_manager/
        __init__.py
        _model_loader.py
        _model_management.py
``` 
and in the model_manager's init file, we run
```
_model_manager = ModelManager()
```

Since not all metrics require running models (e.g. the `text_structure` metrics)

src/langcheck/metrics/zh/reference_based_text_quality.py
```diff
@@ -90,14 +90,13 @@ def semantic_similarity(
                                               openai_args)
         metric_value.language = 'zh'
         return metric_value
+    # lazy import
+    from langcheck.metrics import _model_manager
+    model = _model_manager.fetch_model(language='zh',
+                                       metric_type="semantic_similarity")
```
yosukehigashi, line 96:
```suggestion
                                       metric="semantic_similarity")
```

src/langcheck/metrics/zh/reference_free_text_quality.py
```diff
@@ -90,9 +90,17 @@ def sentiment(
...
+        model=_sentiment_model_path)  # type: ignore[reportGeneralTypeIssues]
     # {0:"Negative", 1:'Positive'}
+    from langcheck.metrics import _model_manager
+    tokenizer, model = _model_manager.fetch_model(lanaguage='zh',
+                                                  metric='sentiment')
+    _sentiment_pipeline = pipeline(
+        'sentiment-analysis',
+        model=model,  # type: ignore[reportGeneralTypeIssues]
+        tokenizer=tokenizer  # type: ignore[reportGeneralTypeIssues]
+    )
```
yosukehigashi, line 103: I don't think this is working? If I understand correctly, we should be using the `load_pipeline_for_text_classification` function so that `fetch_model` will directly return a `pipeline` object

Vela-zz: > I don't think this is working? If I understand correctly, we should be using the `load_pipeline_for_text_classification` function so that `fetch_model` will directly return a `pipeline` object

I think using pipeline(model= model, tokenizer=tokenizer), would return the same result with the traditional way that call tokenizer() fisrt and then model(), becuase pipeline help us do this inside its class by preprosee(),_forward() and postprocess(). so the code here would not generate different result. May be we should require all people use the same inference way first.
image

2. That's the point I wonder how to handle it. 

First, I check other language module, and pipeline is used in zh only. So, for current situation, a specific load_pipeline() seems redudant. I modify the code here in zh to make a load_auto_model() fit both pipeline inference way and traditional inference way, make loader become more simple.

------ Other findings when check your comment ------------
if our final goal is letting user modify model sets, unique postprocessing in text-classification seems impossible to be set in configuration. Because the number of class is different across models. Although, most of models may do binary classification task, but like toxicity model in zh, it has 5 class, I use a [1- sum(logits[1:5])] to transform logits to score, and in ja toxicity, use [ logits[0] + 0.5 * logits[1]] postprocessing to handle a 3 class text-classification. 
```
model_manager
  - __init__(): call from_file inside 
  - register(language, metric, model, tokenizer, reivsion, *args, **kwargs)
  - get_model(lang, metric, using_pipeline: bool):
  - from_file(file_name): use register(...) & validate(...) inside
  - validate(): check model availablity 
  - list_models(...)
  - save(): output conifg
```

yosukehigashi: Thanks for the detailed comment! Those are some good points / questions, and definitely things we need to think about eventually. But for now, I think we can focus on just making `fetch_model` work for our existing models.

Basically I was running into some issues before, but a couple of them are fixed now after your commit (https://github.com/citadel-ai/langcheck/pull/73/commits/4433da1afc441ec6d9df8c254223358ec2620deb). I'm still running into a couple other issues with the `factual_consistency` model (I'll add a comment there directly)

src/langcheck/metrics/_model_management.py
```diff
@@ -0,0 +1,219 @@
...
+                if loader_type == 'sentence-transformers':
+                    if revision is not None:
+                        print(
+                            'Info: Sentence-Transformers do not support fixed model versions yet'  # NOQA:E501
+                        )
+                    model = load_sentence_transformers(model_name=model_name)
+                    return model
+                elif loader_type == 'huggingface':
+                    tokenizer_name = config.pop('tokenizer_name', None)
+                    tokenizer, model = load_auto_model_for_text_classification(
```
yosukehigashi, line 78: For `factual_consistency`, the model is a text2text model so I think it needs to be downloaded with `AutoModelForSeq2SeqLM.from_pretrained` instead of `AutoModelForSequenceClassification.from_pretrained`. Maybe add another loader in `_model_loader.py`?

src/langcheck/metrics/model_manager/_model_management.py
```diff
@@ -0,0 +1,257 @@
...
+                detail_config['tokenizer_name'] = tokenizer_name
+            # If model's revision is pinned
+            revision = kwargs.pop('model_revision', None)
+            if revision:
+                detail_config['revision'] = revision
+
+            # Validate the change
+            if ModelManager.validate_config(self.config,
+                                            language=language,
+                                            metric=metric):
```
yosukehigashi, line 207: This doesn't return anything (just raises an exception if something goes wrong) so I think we just need to call it?

src/langcheck/metrics/zh/reference_free_text_quality.py
```diff
@@ -90,9 +90,16 @@ def sentiment(
     global _sentiment_model_path
 
     _sentiment_pipeline = pipeline(
-        'sentiment-analysis', model=_sentiment_model_path
-    )  # type: ignore[reportGeneralTypeIssues]  # NOQA: E501
+        'sentiment-analysis',
+        model=_sentiment_model_path)  # type: ignore[reportGeneralTypeIssues]
```
yosukehigashi, line 94: We can remove this now!

src/langcheck/metrics/zh/reference_free_text_quality.py
```diff
@@ -210,9 +217,15 @@ def _toxicity_local(generated_outputs: List[str]) -> List[float]:
     global _toxicity_model_path
```
yosukehigashi, line 217: We can remove this too

src/langcheck/metrics/zh/source_based_text_quality.py
```diff
@@ -86,8 +86,11 @@ def factual_consistency(
 
     global _factual_consistency_translation_pipeline
     if _factual_consistency_translation_pipeline is None:
```
yosukehigashi, line 88: I don't think we need to make this a global var anymore, since the model manager is maintaining it?

yosukehigashi (changes requested):
Thanks for the updates @Vela-zz!! This is looking really good now 😄 

(FYI I made a bunch of cleanup commits, docstrings and comments mostly)

tests/metrics/model_manager/test_model_loader.py
```diff
@@ -31,59 +31,51 @@ def test_load_auto_model_for_seq2seq(model_name, tokenizer_name, revision):
         # Check if the tokenizer was loaded correctly
```
yosukehigashi, line 31: This comment seems a bit out of place now. I think we can delete it?

tests/metrics/model_manager/test_model_loader.py
```diff
@@ -31,59 +31,51 @@ def test_load_auto_model_for_seq2seq(model_name, tokenizer_name, revision):
...
                return_value=MockTokenizer) as mock_tokenizer, \
-         patch('transformers.AutoModelForSequenceClassification.from_pretrained',
+         patch('transformers.AutoModelForSequenceClassification.from_pretrained',  # NOQA:E501
                return_value=MockSeqClassifcationModel) as mock_model:
-        tokenizer, model = load_auto_model_for_text_classification(model_name,
-                                                                   tokenizer_name, revision)  # NOQA:E501
+        tokenizer, model = load_auto_model_for_text_classification(
+            model_name, tokenizer_name, revision)  # NOQA:E501
 
         # Check if the tokenizer was loaded correctly
```
yosukehigashi, line 56: Same thing here, I think we can delete it

tests/metrics/model_manager/test_model_manager.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@pytest.fixture
+def temp_config_path(tmp_path):
+    config = """
+    zh:
+      toxicity:
+        model_name: "Alnusjaponica/toxicity-score-multi-classification"
+        tokenizer_name: "line-corporation/line-distilbert-base-japanese"
+        loader_func: "load_auto_model_for_text_classification"
```
yosukehigashi, line 27: (Just curious) Why not use the params for `zh`? I.e.
```
    model_name: alibaba-pai/pai-bert-base-zh-llm-risk-detection
    model_revision: 0a61c79744cb0173216f015ffecc1ea81c4e0229
    loader_func: load_auto_model_for_text_classification
```

src/langcheck/metrics/model_manager/_model_loader.py
```diff
@@ -0,0 +1,91 @@
...
+        tokenizer: The loaded tokenizer.
+        model: The loaded sequence classification model.
+    '''
+    if tokenizer_name is None:
+        tokenizer_name = model_name
+    tokenizer = AutoTokenizer.from_pretrained(tokenizer_name,
+                                              trust_remote_code=True,
+                                              revision=revision)
+    model = AutoModelForSequenceClassification.from_pretrained(
+        model_name, revision=revision, trust_remote_code=True)
```
yosukehigashi, line 62: Is `trust_remote_code=True` for the current selection of `zh` models?

yosukehigashi (commented):
Basically LGTM! Couple last questions, but looks good to merge after that 😄

src/langcheck/metrics/model_manager/_model_loader.py
Vela-zz: > Is `trust_remote_code=True` for the current selection of `zh` models?
No, but I copied part of ja config [ja - toxicity] for test, and it seems I need add trust_remote_code to let the model load properly. so I add it.

yosukehigashi: Ah ok, then maybe we can change the test to run on the `zh` config instead and remove `trust_remote_code=True` for now? The Japanese models are models finetuned by our team, so we can try and change it to not require `trust_remote_code=True` when we integrate the Model Manager with Japanese models

yosukehigashi (approved):
LGTM, thanks for all of your work on this @Vela-zz!! I'll merge this to the zh branch once tests pass 🎉

PR #76, opened 2023-12-14
Metric to compute context relevance

yosukehigashi:
Adds a metric to compute how relevant the retrieved source (i.e. context) is to the user's prompt by asking OpenAI. The English and Japanese versions have different prompts, so that the explanations will be written in the appropriate language.


liwii (approved):
At least both `ja` and `en` metric worked well in my environment.

LGTM, it looks very useful!!

PR #77, opened 2023-12-26
Metric to compute answer relevance

yosukehigashi:
Adds a metric to compute how relevant the generated output is to the user's prompt by asking OpenAI. The English and Japanese versions have different prompts, so that the explanations will be written in the appropriate language.

This is related to https://github.com/citadel-ai/langcheck/pull/76, but measure's the answer's relevance to the prompt instead of the retrieved context's to the prompt.


liwii (approved):
Confirmed both `en` and `ja` work well. LGTM.

PR #81, opened 2024-01-27
German improvements

ischender:
Tiny changes:
a) adding DE to the main README
b) adding a notebook for factual consistency (it was still work in progress at the time of last PR)

PR #82, opened 2024-01-27
German prompts

ischender:
Moving most of the functions involving openAI to the German code, and translating prompts.

### DIFF
--- a/src/langcheck/metrics/de/reference_free_text_quality.py
+++ b/src/langcheck/metrics/de/reference_free_text_quality.py
@@ -230,11 +230,47 @@ def fluency(
     openai_client: Optional[OpenAI] = None,
     openai_args: Optional[Dict[str,
                                str]] = None) -> MetricValue[Optional[float]]:
-    ''' If using the local models, we first translate the generated outputs
-    to English, and then use the Parrot fluency model to calculate the fluency
-    scores, from the English counterpart. If using the OpenAI models, we use
-    the OpenAI API to calculate the fluency scores. This metric takes on float
+    ''' Calculates the fluency scores of generated outputs. This metric takes on
+    float values between [0, 1], where 0 is low fluency and 1 is high fluency.
+    (NOTE: when using the OpenAI model, the fluency scores are either 0.0
+    (poor), 0.5 (fair), or 1.0 (good). The score may also be `None` if it could
+    not be computed.)
+
+    We currently support three model types:
+
+    1. The 'local' type, we first translate the generated outputs
+    to English, then use the Parrot fluency model for the English counterpart.
+    This is the default model type and there is no setup needed to run this.
+
+    2. The 'openai' type, where we use OpenAI's 'gpt-turbo-3.5' model
+    by default. While the model you use is configurable, please make sure to use
+    one that supports function calling
+    (https://platform.openai.com/docs/guides/gpt/function-calling). See
+    `this page <https://langcheck.readthedocs.io/en/latest/metrics.html
+    #computing-metrics-with-openai-models>`__
+    for examples on setting up the OpenAI API key.
+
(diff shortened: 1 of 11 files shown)


src/langcheck/metrics/de/reference_free_text_quality.py
```diff
@@ -97,44 +95,135 @@ def sentiment(
     # The English prompt works well enough for German
     # TODO: Investigate the performance improvement with German prompt
```
yosukehigashi, line 96: This is done!

src/langcheck/metrics/de/reference_free_text_quality.py
```diff
@@ -97,44 +95,135 @@ def sentiment(
...
+    explanation_list = []
+    for gen in tqdm_wrapper(generated_outputs):
+        score, explanation = oai_evaluator.get_score(_prompt(gen_output=gen),
+                                                     _function_call_prompt)
+        score_list.append(score)
+        explanation_list.append(explanation)
+    return score_list, explanation_list
+
+
 def fluency(
```
yosukehigashi, line 227: Fluency also has an OpenAI version, so for that one, we can pass in the German text directly.

src/langcheck/metrics/de/source_based_text_quality.py
```diff
@@ -83,34 +83,139 @@ def factual_consistency(
     # The English prompt works well enough for German, like with Japanese
     # TODO: Investigate performance improvement with German prompt / translation
     if model_type == 'openai' or model_type == 'azure_openai':
```
yosukehigashi, line 85: This is done!

src/langcheck/metrics/de/reference_free_text_quality.py
yosukehigashi, line 1: We recently added a new metric called `answer_relevance` in the English and Japanese versions, which just has an OpenAI version. So if you could also translate the prompt to german, that would be great!

yosukehigashi (commented):
Thanks @ischender!

src/langcheck/metrics/de/reference_free_text_quality.py
ischender: True, to be honest, I did not investigate the improvements...

src/langcheck/metrics/de/reference_free_text_quality.py
ischender: ah yes, I missed that, adding that too...

yosukehigashi (commented):
Thanks! Just a couple of docstring comments, but otherwise LGTM. I'll take a final look after #81 is merged first

src/langcheck/metrics/de/reference_free_text_quality.py
```diff
@@ -94,78 +93,273 @@ def sentiment(
...
     openai_client: Optional[OpenAI] = None,
     openai_args: Optional[Dict[str,
                                str]] = None) -> MetricValue[Optional[float]]:
-    ''' We first translate the generated outputs to English, and then use the
-    Parrot fluency model to calculate the fluency scores, from the English
-    counterpart.
+    ''' If using the local models, we first translate the generated outputs
+    to English, and then use the Parrot fluency model to calculate the fluency
+    scores, from the English counterpart. If using the OpenAI models, we use
+    the OpenAI API to calculate the fluency scores. This metric takes on float
```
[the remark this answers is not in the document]
ischender: good point, done it now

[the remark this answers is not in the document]
yosukehigashi: Thanks!

yosukehigashi (approved):
LGTM! Excited to try this out on some German text (even though I won't understand it 😆). I'll merge this once tests pass!

PR #87, opened 2024-02-19
Japanese prompts for reference free OpenAI-based metrics

yosukehigashi:
Updates all reference free OpenAI-based metrics to Japanese prompts.

Updating the factual consistency prompt to Japanese is still a TODO - this one was more difficult and all our attempts considerably lowered the performance on the JNLI benchmark.


src/langcheck/metrics/ja/reference_free_text_quality.py
```diff
@@ -122,15 +142,92 @@ def sentiment(
...
+            ``client.chat.completions.create`` function
+
+    Returns:
+        score_list: a list of scores
+        explanation_list: a list of explanations for the scores
+    '''
+
+    def _prompt(gen_output: str) -> str:
+        return f'''
+        提出されたテキストのセンチメントを評価してください。データは以下の通りです:
```
liwii, line 179: "センチメント" doesn't seem familiar to me even as a katakana-word. Maybe "感情" instead?

liwii (approved):
Used it for a few times and they seem working fine.

Left just one nit comment.

src/langcheck/metrics/ja/reference_free_text_quality.py
liwii: LGTM!!

PR #93, opened 2024-02-29
Move nltk download to inside the gender augmentation function

yosukehigashi:
This ensures that `nltk.download` will only run when `augment.gender` is actually run, rather than when `import langcheck` is run.

### DIFF
--- a/src/langcheck/augment/en/_gender/_gender.py
+++ b/src/langcheck/augment/en/_gender/_gender.py
@@ -72,7 +72,7 @@ def _replace_gender_pronouns(
         str: Augmented text.
     """
     try:
-        nltk.data.find('averaged_perceptron_tagger')
+        nltk.data.find('taggers/averaged_perceptron_tagger')
     except LookupError:
         nltk.download('averaged_perceptron_tagger')
     try:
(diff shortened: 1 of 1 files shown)


yosukehigashi:
> LGTM for the changes, let me suggest one more thing for this PR:
> 
> Found that actually the right patyh for `nltk.data.find` and `nltk.download` are different. That's why the code always takes `except LookupError` and shows the `nltk` log like that.
> 
> ```
> [nltk_data] Downloading package averaged_perceptron_tagger to
> [nltk_data]     /home/vscode/nltk_data...
> [nltk_data]   Package averaged_perceptron_tagger is already up-to-
> [nltk_data]       date!
> [nltk_data] Downloading package punkt to /home/vscode/nltk_data...
> [nltk_data]   Package punkt is already up-to-date!```
> ```
> 
> [nltk_data] Downloading package averaged_perceptron_tagger to [nltk_data] /home/vscode/nltk_data... [nltk_data] Package averaged_perceptron_tagger is already up-to- [nltk_data] date! [nltk_data] Downloading package punkt to /home/vscode/nltk_data... [nltk_data] Package punkt is already up-to-date!
> 
> ```
> 
> Could you double-check that my two suggestions resolve the problem?
> ```

ohhh yeah good find!! We were doing this properly for stats and factual consistency 😅 Checked that your suggestions work

liwii (approved):

PR #94, opened 2024-03-02
Refactoring of the local metrics

liwii:
Refactored (mainly) local metrics code. The changes include:

- For reference_free metrics with `Detoxify` / `AutoModelForSequenceClassification`, created base classes and made logics such as `tqdm` progress bar common.  For these metrics I also added an argument `local_overflow_strategy`, with which you can handle the input longer than the model input window in 3 different ways (`truncate`, `nullify`, `raise`)
- For semantic similarity (both local and openai), made a base class and put most of the logic into that class.
- For non-english factual consisntency, long inputs always caused error, so avoided it by adding `truncation=True` argument. (I would like to add the overflow handling to the factual consietency metrics too, but it should be huge enough to develop in another PR)
- Moved loading logics of many models into `model_manager`. `check_model_availability` became flaky as a result, so I disabled these checks for now.


liwii:
@yosukehigashi 
The PR became much bigger than I thought, but I would be happy if you could take a look when you have time!!

src/langcheck/metrics/en/reference_based_text_quality.py
```diff
@@ -70,96 +70,18 @@ def semantic_similarity(
...
-                openai_client = AzureOpenAI(
-                    api_key=os.getenv("AZURE_OPENAI_KEY"),
-                    api_version=os.getenv("OPENAI_API_VERSION"),
-                    azure_endpoint=os.getenv(
-                        "AZURE_OPENAI_ENDPOINT"))  # type: ignore
-        if model_type == 'azure_openai' and not openai_args:
-            raise AssertionError(
-                'The embedding model deployment must be specified in '
-                '`openai_args` for the azure_openai type, e.g. '
-                '`openai_args={"model": "YOUR_DEPLOYMENT_NAME"}`')
```
yosukehigashi, line 109: I think we should keep this check?

src/langcheck/metrics/model_manager/_model_management.py
```diff
@@ -144,22 +145,25 @@ def validate_config(config, language='all', metric='all') -> None:
...
-                # Check tokenizer availability with revision if specified.
-                tokenizer_name = model_setting.get('tokenizer_name')
-                if tokenizer_name is not None and tokenizer_name != model_name:
-                    tokenizer_revision = model_setting.get('tokenizer_revision')
-                    if not check_model_availability(tokenizer_name,
-                                                    tokenizer_revision):
-                        raise ValueError(
-                            f'Cannot find {tokenizer_name} with {tokenizer_revision} ay Huggingface Hub'  # NOQA:E501
-                        )
+                # TODO: Check the model and tokenizer availability
```
yosukehigashi, line 148: Instead of commenting it out, maybe we can add a param (e.g. `run_check_model_availability`) to `validate_config`, set it to False by default, and add an `if run_check_model_availability` statement here? That way we won't run it during tests, but users can check availability if they want to

src/langcheck/metrics/de/reference_based_text_quality.py
```diff
@@ -80,63 +79,17 @@ def semantic_similarity(
...
-            # can cause the dot product of almost identical
-            # vectors to exceed 1.0 slightly,
-            # so we clip the outputs
-            cosine_scores = torch.clamp(cosine_scores, -1.0, 1.0)
-            scores.extend(cosine_scores.tolist())
+    if model_type == 'local':
+        # https://www.sbert.net/docs/pretrained_models.html#multi-lingual-models
+        # v1 supports only 15 languages (German included) but is stronger than
+        # v2 that supports 50+ languages
+        # NOTE: it's cased! "Das ist ein Test." != "das ist ein test."
```
yosukehigashi, line 86: maybe move this comment to the metric config yaml file?

src/langcheck/metrics/en/reference_based_text_quality.py
```diff
@@ -70,96 +70,18 @@ def semantic_similarity(
         'local', 'openai', 'azure_openai'
     ], ('Unsupported embedding model type. '
         'The supported ones are ["local", "openai", "azure_openai"]')
-    batch_size = 8
     if model_type == 'local':
         # The 'all-mpnet-base-v2' model has the highest average performance out
         # of all the existing sentence-transformer models that have been
         # evaluated.
         # Ref: https://www.sbert.net/docs/pretrained_models.html#model-overview
```
yosukehigashi, line 77: move this to the yaml file?

src/langcheck/metrics/ja/reference_free_text_quality.py
```diff
@@ -116,40 +107,19 @@ def _sentiment_local(generated_outputs: List[str]) -> List[float]:
...
-            probs = torch.nn.functional.softmax(
-                _sentiment_model(**batch_input_tokens).logits, dim=1)
-            scores.extend((probs[:, 1] / 2 + probs[:, 2]).tolist())
-    return scores
+    scorer = AutoModelForSequenceClassificationScorer(
+        language='ja',
+        metric='sentiment',
+        class_weights=[0, 0.5, 1],
+        overflow_strategy=overflow_strategy,
+        max_input_length=512)
```
yosukehigashi, line 121: (Just curious) why do we sometimes need to specify this manually? Does this not have a `config.max_position_embeddings`?

yosukehigashi (commented):
Wow what a beautiful refactor 🤩 Left a bunch of comments but generally LGTM!

Let's do a sanity check with LangCheckChat as well since this is a pretty big change. Could you test out English and Japanese, and I'll do Chinese and German?

liwii:
Thanks for the review! Let me resolve the comments first, then we can do the sanity-check together as you say!

src/langcheck/metrics/scorer/_base.py
```diff
@@ -0,0 +1,134 @@
...
+from sentence_transformers import util
+from torch import Tensor
+
+from langcheck.utils.progess_bar import tqdm_wrapper
+
+# Define a type variable for token type.
+# This type is used to represent the list of tokens returned by the
+# _tokenize method. We do not use `list` type because the token type
+# can be a list, dict, or any other type.
+_TokenType = TypeVar('_TokenType')
```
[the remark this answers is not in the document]
liwii: Makes sense, renamed the type variable.

src/langcheck/metrics/scorer/_base.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class BaseSingleScorer(Generic[_TokenType]):
+    '''Base class for single input scorers.
+    '''
+    BASE_BATCH_SIZE = 8
```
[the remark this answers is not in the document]
liwii: Ah yeah I wanted it to be overridable. Just updated the code to make it an instance variable `self.batch_size`.

src/langcheck/metrics/de/_translation.py
```diff
@@ -16,7 +16,8 @@ def __init__(self, model_name: str) -> None:
         '''
         self._translation_pipeline = pipeline("translation",
                                               model=model_name,
-                                              tokenizer=model_name)
+                                              tokenizer=model_name,
+                                              truncation=True)
         self._max_length = self._translation_pipeline.model.config.max_length
 
     def _translate(self, texts: str) -> str:
```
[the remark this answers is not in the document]
liwii: So basically the minimum unit of the split text is a sentence, so if one sentence contains longer input than 512 tokens, it fails in the last pipeline calling part.

The `truncation=True` is only applied when you directly call the pipeline, and it doesn't affect the `self._translation_pipeline.tokenizer` call, so it should be fine.

src/langcheck/metrics/ja/reference_free_text_quality.py
liwii: Seemingly some models are not accurate about their `config.max_position_embeddings`. For example, this model has `max_position_embeddings=514` but it doesn't seem to accept input tokens longer than 514.
https://huggingface.co/cardiffnlp/twitter-xlm-roberta-base-sentiment-multilingual/blob/main/config.json

There are other models that do not define `max_input_length` too, that's why I made it one of the init arguments.

src/langcheck/metrics/scorer/hf_models.py
```diff
@@ -0,0 +1,131 @@
...
+from langcheck._handle_logs import _handle_logging_level
+
+from ._base import BaseSimilarityScorer, BaseSingleScorer
+
+
+class AutoModelForSequenceClassificationScorer(BaseSingleScorer):
+    '''Scorer using Hugging Face's AutoModelForSequenceClassification.
+    '''
+
+    def __init__(self,
```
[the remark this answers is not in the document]
liwii: Added a simple docstring there

src/langcheck/metrics/scorer/hf_models.py
```diff
@@ -0,0 +1,131 @@
...
+        return (
+            {
+                key: value[start_idx:end_idx]
+                for key, value in input_tokens.items()
+            },  # type: ignore
+            validation_results[start_idx:end_idx])
+
+    def _logits_to_scores(self, logits: torch.Tensor) -> list[float]:
+        '''Turn the logits returned from the models to scores.
+        The users can override this method to customize the behavior.
```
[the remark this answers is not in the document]
liwii: Ah at the beginning there was no `class_weights` so I was thinking of overriding `_logits_to_scores` to apply it to different types of models. Turned out `class_weights` is enough for the original purpose. Deleted the line.

src/langcheck/metrics/en/reference_based_text_quality.py
liwii: Removed it totally by accident. Thanks for the good catch.

src/langcheck/metrics/model_manager/_model_management.py
liwii: Sounds good, fixed the code in that way

liwii:
@yosukehigashi 
Does the change look good to you? If so, we can start testing the new version.

yosukehigashi:
> @yosukehigashi Does the change look good to you? If so, we can start testing the new version.

yeah looks fine to me! can you start testing English and Japanese then? I need to make some code changes to test German / Chinese

liwii:
Tried some examples of Japanese & English with references, there were no errors at least.

yosukehigashi (approved):
Chinese and German work fine too. LGTM!

PR #95, opened 2024-03-04
Add a tutorial for LangCheckChat

yosukehigashi:
Tutorial based on LangCheckChat!


docs/tutorial_langcheckchat.md
```diff
@@ -0,0 +1,125 @@
+# LangCheckChat: a Q&A system over LangCheck Docs that Auto-Evaluates itself
+
+## Introduction
+
+In this tutorial, we will explore how to build a very simple RAG system (which we call LangCheckChat) that allows you to ask questions about the [LangCheck documentation](https://langcheck.readthedocs.io/en/latest/). Then, we will explore how we can use LangCheck metrics to evaluate the system’s performance, and show those results to the user after each response.
+
+[LangCheckChat](https://github.com/citadel-ai/langcheckchat) is fully open source, so please follow along by trying it yourself if you can!
+
```
kennysong, line 8: I think it's good to have a `## Video Demo` section here with the demo video to capture people's attention early

docs/tutorial_langcheckchat.md
```diff
@@ -0,0 +1,125 @@
...
+### Step 4: (Optional) Check the reference-based metric results
+If you entered a reference answer as outlined in step 3, you should see a new metrics table called [Reference-Based Text Quality Metrics](https://langcheck.readthedocs.io/en/latest/langcheck.metrics.en.reference_based_text_quality.html). These metrics compute how similar the LLM’s answer is to the reference text in various ways.
+
+![LangCheckChat 4](_static/LangCheckChat_4.png)
+
+### Step 5: (Optional) Check the logs of past interactions
+At the bottom of the page, there’s a link that says “See Q&A Logs”, and clicking that will take you to the logs page. You should see your latest interaction with LangCheckChat in the logs table, and all future interactions will be similarly tracked in this table.
+
+![LangCheckChat 5](_static/LangCheckChat_5.png)
```
kennysong, line 125: Might be good to have screenshots of a hallucination or otherwise bad answer and link to the code that runs the metric and adds the warning

docs/tutorial_langcheckchat.md
```diff
@@ -0,0 +1,125 @@
...
+
+Here’s the response from the LLM:
+> To install LangCheck, you can run the following command:
+>
+> pip install langcheck
+>
+> Please note that LangCheck requires Python 3.8 or higher to work properly.
+
+We can also see the sources that were retrieved from the index. By default, the top 2 most relevant source nodes are returned, which is what we see in `response.source_nodes`.
+
```
kennysong, line 69: Maybe add sample code here to explain how to use LangCheck to measure ai_disclaimer and factual_consistency?

kennysong (commented):
Reviewed! This is very well written and easy to follow. I think we need to add a few more examples to make it more of a LangCheck tutorial though.

Also pushed a commit to actually include the tutorial onto the tutorials page

docs/tutorial_langcheckchat.md
kennysong: Pushed a commit that installs a sphinx extension so we can embed YouTube videos. Do you have an account for uploading the video?

We can also use https://github.com/sphinx-contrib/video but don't want to include the .mp4 inside the repo and I'm not sure how to store static files in GitHub

yosukehigashi: hmm is what I did for the demo page in langcheckchat not good? https://github.com/citadel-ai/langcheckchat?tab=readme-ov-file#demo

I just uploaded as a github asset

docs/tutorial_langcheckchat.md
yosukehigashi: Added a screenshot in https://github.com/citadel-ai/langcheck/pull/95/commits/dc588e55d0c9d1b39fce89a31c4a163d76ef16ee

docs/tutorial_langcheckchat.md
yosukehigashi: I added a bunch of sample code in https://github.com/citadel-ai/langcheck/pull/95/commits/dc588e55d0c9d1b39fce89a31c4a163d76ef16ee, ptal!

docs/tutorial_langcheckchat.md
kennysong: This markdown file is not rendered by GitHub, it's converted by Sphinx into HTML. So to embed a raw mp4 file, we need to use https://github.com/sphinx-contrib/video

But to do this we need to point to the mp4 file in this repo, or have a URL that ends in .mp4. 

The URL in https://github.com/citadel-ai/langcheckchat?tab=readme-ov-file#demo doesn't end in .mp4 so I couldn't use it directly (I'm also not sure how you uploaded it since it's not a file in the repo)

yosukehigashi: Ahhh I see. In langcheckchat I just did a drag and drop, similar to how you can upload videos to PR descriptions.

Do we have a company YouTube account that we can use?

(Also, OOC why is uploading the mp4 file to the repo a problem (assuming we don't hit the large file limit)?)

docs/tutorial_langcheckchat.md
kennysong: Thanks! I added a sentence linking to the actual threshold since it's a useful reference

docs/tutorial_langcheckchat.md
kennysong: Ah just remembered we have a Citadel AI Vimeo account! I uploaded it there. The video player is a lot cleaner than YouTube too

(The video file is 50MB so would rather not keep it in the core repo unless there's no other way)

yosukehigashi: oh cool, thanks!! so just to confirm, the video will show up in readthedocs but is not expected to render in GitHub? (This is what I'm seeing right now)
image

kennysong: Yep, it should show up on readthedocs but not in GitHub. You can generate the readthedocs UI locally by running step 2 and 3 here: https://langcheck.readthedocs.io/en/latest/contributing.html#documentation

yosukehigashi: thanks! checked that it's showing up in readthedocs

Is it fine for it to not render properly in GitHub though? I guess we already have some sections that render better in readthedocs

kennysong: Yeah, I think the docs in GitHub are raw files and aren't meant to be read directly. (e.g. most of the docs/ folder are .rst files which don't contain any information before they're auto-populated when Sphinx converts them into HTML)

PR #101, opened 2024-03-19
Change input of `OpenAIBasedEvaluator.get_score()` from str to `Iterable[str]`

Alnusjaponica:
## Motivation
First step to enable async API call of OpenAI API.

## Details
This PR allows `OpenAIBasedEvaluator.get_score()` to take Iterable[str] in addition to `str`, which enables metrics functions to pass prompt at once.

### DIFF
--- a/src/langcheck/metrics/en/_openai.py
+++ b/src/langcheck/metrics/en/_openai.py
@@ -100,7 +100,6 @@ class OpenAIBasedEvaluator:
         # First, call the API to get an unstructured assessment. The response
         # should include both the assessment itself and the model's reasoning
         # for that assessment.
-        # Call the API to get an unstructured assessments.
         config_unstructured_assessments = {
             "model": "gpt-3.5-turbo",
             "seed": 123
@@ -168,8 +167,7 @@ class OpenAIBasedEvaluator:
                                         in self._assessment_to_score_mapping):
                 continue
             # By leveraging the function calling API, this should be pretty
-            # rare, but we're dealing with LLMs here so nothing is
-            # absolute!
+            # rare, but we're dealing with LLMs here so nothing is absolute!
             print(f'OpenAI returned an unrecognized assessment: "{assessment}"')
 
         return [
--- a/src/langcheck/metrics/en/_openai.py
+++ b/src/langcheck/metrics/en/_openai.py
@@ -6,7 +6,8 @@ from collections.abc import Callable, Iterable
 from typing import Any
 
 from openai import AzureOpenAI, OpenAI
-from tqdm import tqdm
+
+from langcheck.utils.progess_bar import tqdm_wrapper
 
 
 class OpenAIBasedEvaluator:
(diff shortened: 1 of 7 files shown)


Alnusjaponica:
@liwii I split this change from #89 to make review easier. Could you take a look?

src/langcheck/metrics/en/_openai.py
```diff
@@ -129,38 +139,72 @@ def get_score(
...
+                return None
+            try:
+                return self._client.chat.completions.create(**model_input)
+            except Exception as e:
+                return e
+
+        model_inputs = map(_generate_model_input, prompts)
+        responses = [
+            _call_api_with_exception_filter(model_input)
+            for model_input in tqdm(model_inputs)
```
liwii, line 200: For all the other places, `langcheck.utils.progess_bar.tqdm_wrapper` (I just found the typo here) is used to show the progress bar. Could you follow that here too for consistency?

If the reason why you used `tqdm` here is that the `tqdm_wrapper` exhausts the `map` object by getting the `total=` value inside, I think we should provide the `total=` value somehow from the outside. Otherwise we cannot show the progress bar, and that is very important for the user experience.

liwii (commented):
Sorry for the pretty late review! And thanks for the nice refactoring!!

Left a few nit comments and one comment about the tqdm progress bar.

src/langcheck/metrics/en/_openai.py
```diff
@@ -129,38 +139,72 @@ def get_score(
...
+        ]
+
+        # Check if any of the assessments are not recognized.
+        for assessment in assessments:
+            if (assessment is None) or (assessment
+                                        in self._assessment_to_score_mapping):
+                continue
             # By leveraging the function calling API, this should be pretty
-            # rare, but we're dealing with LLMs here so nothing is absolute!
+            # rare, but we're dealing with LLMs here so nothing is
```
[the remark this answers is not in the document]
Alnusjaponica: Thanks! I updated

src/langcheck/metrics/en/_openai.py
```diff
@@ -73,46 +79,50 @@ def get_score(
...
-                response = self._client.chat.completions.create(
-                    messages=messages,  # type: ignore
-                    seed=123,
-                    **self._openai_args)  # type: ignore
-            unstructured_assessment = response.choices[0].message.content
-        except Exception as e:
-            print(f'OpenAI failed to return an unstructured assessment: {e}')
-            print(f'Prompt that triggered the failure is:\n{prompt}')
-            return None, None
+        # Call the API to get an unstructured assessments.
```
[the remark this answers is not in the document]
Alnusjaponica: Thanks! I updated

src/langcheck/metrics/en/_openai.py
Alnusjaponica: Sure, I updated the code to convert the input into `list` and use  `langcheck.utils.progess_bar.tqdm_wrapper`.

liwii: Hmm it actually does not work because this line "exhausts" all the contents of the iterator and actual `tqdm` gets an empty iterator.
https://github.com/Alnusjaponica/langcheck/blob/9eb2e49286d1228a11b95dbfaf56180a2b063310/src/langcheck/utils/progess_bar.py#L16

Let's think about a good way to avoid that

liwii (approved):
LGTM!!

PR #106, opened 2024-04-01
Update pytest status badge and run tests on PR merge

kennysong:
Based on IRL conversation. We now run pytest when a PR merges into main, and the status badge is updated to only read these events.

It's a bit wasteful but I didn't find an alternative way to do this. 

Relevant documentation: https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge#using-the-event-parameter


yosukehigashi (approved):
Oh nice!! LGTM

PR #108, opened 2024-04-03
Japanese pairwise comparison metric

yosukehigashi:
- Refactored language agnostic logic of the `pairwise_comparison` metric into `src/langcheck/metrics/_pairwise_text_quality_utils.py`
- Implemented `pairwise_comparison` for `ja` with prompts in Japanese
- Enabled async for `pairwise_comparison` (both en and ja)

### DIFF
--- a/src/langcheck/metrics/_pairwise_text_quality_utils.py
+++ b/src/langcheck/metrics/_pairwise_text_quality_utils.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import Callable
 
 
--- a/src/langcheck/metrics/_pairwise_text_quality_utils.py
+++ b/src/langcheck/metrics/_pairwise_text_quality_utils.py
@@ -1,0 +1,59 @@
+from typing import Callable, Optional
+
+
+class PairwiseComparisonPromptGenerator:
+
+    def __init__(self, base_prompt_template: Callable,
+                 prompt_template_with_reference: Callable,
+                 prompt_template_with_source: Callable,
+                 prompt_template_with_source_and_reference: Callable) -> None:
+        self.base_prompt_template = base_prompt_template
+        self.prompt_template_with_reference = prompt_template_with_reference
+        self.prompt_template_with_source = prompt_template_with_source
+        self.prompt_template_with_source_and_reference = prompt_template_with_source_and_reference  # NOQA: E501
+
+    def generate_prompts(self, generated_outputs_1: list[str],
+                         generated_outputs_2: list[str], prompts: list[str],
+                         sources_1: Optional[list[str]],
+                         sources_2: Optional[list[str]],
+                         reference_outputs: Optional[list[str]]) -> list[str]:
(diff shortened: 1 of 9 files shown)


src/langcheck/metrics/_pairwise_text_quality_utils.py
```diff
@@ -0,0 +1,95 @@
+from __future__ import annotations
+
+from typing import Callable
+
+
+class PairwiseComparisonPromptGenerator:
+
+    def __init__(self, base_prompt_template: Callable,
```
liwii, line 8: It would be better if we can show the right arguments & return type of these template functions by either docstring or type annotations.

src/langcheck/metrics/ja/pairwise_text_quality.py
```diff
@@ -0,0 +1,256 @@
...
+    for examples on setting up the OpenAI API key.
+
+    2. The 'azure_openai' type. Essentially the same as the 'openai' type,
+    except that it uses the AzureOpenAI client. Note that you must specify your
+    model deployment to use in ``openai_args``, e.g.
+    ``openai_args={'model': 'YOUR_DEPLOYMENT_NAME'}``
+
+    Ref:
+        Our prompt is similar to the prompt used in
+        https://arxiv.org/abs/2306.05685
```
liwii, line 49: Do we need to mention this for the Japanese prompt too?

src/langcheck/metrics/ja/pairwise_text_quality.py
```diff
@@ -0,0 +1,256 @@
...
+        _prompt_with_source_and_reference)
+    all_prompts = prompt_generator.generate_prompts(generated_outputs_a,
+                                                    generated_outputs_b,
+                                                    prompts, sources_a,
+                                                    sources_b,
+                                                    reference_outputs)
+    scores, explanations = oai_evaluator.get_score(all_prompts,
+                                                   _function_call_prompt)
+
+    if enforce_consistency:
```
liwii, line 238: We might want to show something to the user before moving into the consistency check. When I see the progress bars twice in for the first time I was a bit confused.

liwii (approved):
Mostly LGTM, left a few nit comments.

src/langcheck/metrics/_pairwise_text_quality_utils.py
yosukehigashi: added both type hints and a docstring!

src/langcheck/metrics/ja/pairwise_text_quality.py
yosukehigashi: I made it so that the tqdm descriptions can be overridden, so now it looks like this. WDYT?
image

(I tried just adding a print statement at first, but then it's a bit weird for async where the progress bars don't show up right now)

liwii: Got it, it looks reasonable to me!!

PR #110, opened 2024-04-15
Refactoring LLM-based metrics

liwii:
## Background
There are increasing number of openai-related arguments to each metric function (e.g.,  `openai_args`, `use_async`, ... etc). They are basically metric-agnostic, so changes on these configs practically do not need any updates on the metric functions. However, due to the current interface, we need to update all the functions every time. If we increase the number of evaluation models (like Gemini) and the number of supported metrics, it is getting more and more painful.

To resolve this, we introduce a new `EvalClient` interface and inject the external dependencies from outside of the metrics with an `EvalClient` instance. Some other parts of openai metrics are also refactored (e.g. convert the prompt functions into Jinjia templates)

## Things changed

- Add an `EvalClient` class. Now all the configs related to the LLMs used for the evaluation go to the subclasses of this, and the metric functions will take an `EvalClient` instance as an argument if needed.
- `model_type` argument of all the metric functions are renamed to `eval_model` and starts taking an `EvalClient` instance. It breaks the backward compatibility of the metric functions.
- Create a `prompts/` directory, and move all the prompts used for the evaluation to the directory as Jinjia templates
- Make the prompt used for the score extraction step common for all the metrics
- Updated the options in the German metrics to English ones (e.g. "Good") because the original German options contained some non-ascii characters and they didn't work with prompt-calling

## Things not changed
- `langcheck.augment.rephrase` keeps using the older interface, since
  - I couldn't come up with a simple abstract interface that describes "Run the LLM multiple times with different seeds"
  - Presumably the number of users of `langcheck.augment` is much fewer than that of `langcheck.metrics`
- There is also `scorer/` directly, which tries to achieve the similar thing as `eval_clients` does. I keep these two directories in the repository, but it'll be a future TODO to integrate these two directories.


liwii:
@yosukehigashi 
I just updated `ja.sentiment` based on the new structure. Could you check if it looks good to you? After the design is finalized I'll roll it out to all the other metric functions.

src/langcheck/metrics/ja/reference_free_text_quality.py
```diff
@@ -8,44 +8,37 @@
...
-              openai_args: Optional[Dict[str, str]] = None,
-              local_overflow_strategy: str = 'truncate',
-              *,
-              use_async: bool = False) -> MetricValue[Optional[float]]:
+
+def sentiment(
+        generated_outputs: List[str] | str,
+        prompts: Optional[List[str] | str] = None,
+        model_type: str = 'local',
+        llm_client: Optional[EvalClient] = None,
```
yosukehigashi, line 24: Instead of having both of these, maybe we can simplify this down to
```suggestion
        eval_client: str | EvalClient = 'local',
```

src/langcheck/metrics/ja/reference_free_text_quality.py
```diff
@@ -56,38 +49,33 @@ def sentiment(generated_outputs: List[str] | str,
             optional metadata and not used to calculate the metric.
         model_type: The type of model to use ('local', 'openai', or
             'azure_openai'), default 'local'
-        openai_client: OpenAI or AzureOpenAI client, default None. If this is
-            None but ``model_type`` is 'openai' or 'azure_openai', we will
-            attempt to create a default client.
-        openai_args: Dict of additional args to pass in to the
-            ``client.chat.completions.create`` function, default None
+        llm_client: EvalClient, default None. If this is None but ``model_type``
+            is 'llm', we will attempt to create a default client.
```
yosukehigashi, line 53: This is no longer the case, right?

yosukehigashi (commented):
The overall structure looks great to me! Just one comment

src/langcheck/metrics/ja/reference_free_text_quality.py
liwii: Oh yeah looks good!

I renamed the argument to `eval_model` to make it clear that the argument doesn't only take `EvalClient` here.

liwii:
Thanks! I also updated the tests so that they are generalized to all `EvalClient`s. I'll add the test for `OpenAIEvalClient` later.

src/langcheck/metrics/prompts/en/metrics/pairwise_comparison.j2
yosukehigashi, line 1: I feel like this would be easier to read if were organized like:

```
{% if src is not None and ref_output is not None %}
...
{% elif src is not None %}
...
{% elif ref_output is not None %}
...
{% else %}
...
{% endif %}
```

There'll be a lot of duplication, but IMO being able to read the prompt easily is more important

src/langcheck/metrics/zh/reference_based_text_quality.py
```diff
@@ -39,57 +36,36 @@ def semantic_similarity(
...
-    by default (this is configurable). See
-    `this example <https://langcheck.readthedocs.io/en/latest/metrics.html
-    #computing-metrics-with-openai-models>`__
-    for examples on setting up the OpenAI API key.
-
-    3. The 'azure_openai' type. Essentially the same as the 'openai' type,
-    except that it uses the AzureOpenAI client. Note that you must specify your
-    model deployment to use in ``openai_args``, e.g.
-    ``openai_args={'model': 'YOUR_DEPLOYMENT_NAME'}``
+    2. The EvalClient type, where you can use a similarlity scorer returned by
```
yosukehigashi, line 39: Need to update `We currently support three embedding model types:` to `two` above

yosukehigashi (changes requested):
Mostly done with the review (need to take a closer look at the tests, but from a quick glance they lgtm)!!

PTAL! (Most of the comments are just nits on comments / docstrings)

tests/metrics/eval_clients/_openai.py
```diff
@@ -0,0 +1,111 @@
...
+
+from openai.types.chat import ChatCompletion
+
+from langcheck.metrics.eval_clients import (AzureOpenAIEvalClient,
+                                            OpenAIEvalClient)
+
+
+def test_get_text_response_openai():
+    prompts = ['What is the capital of France?'] * 2
+    answer = 'Paris is the capital of France.'
```
yosukehigashi, line 15:
```suggestion
    prompts = ['Assess the factual consistency of the generated output...'] * 2
    answer = 'The output is fully factually consistent.'
```
nit: just to make the example a bit more consistent with what the prompts and answer will actually be like

tests/metrics/eval_clients/_openai.py
```diff
@@ -0,0 +1,111 @@
...
+        assert len(responses) == len(prompts)
+        for response in responses:
+            assert response == answer
+
+
+def test_get_float_score_openai():
+    unstructured_assessment_result: list[str | None] = [
+        'Paris is the capital of France.'
+    ] * 2
+    short_assessment_result = 'Paris'
```
yosukehigashi, line 36:
```suggestion
    unstructured_assessment_result: list[str | None] = [
        'The output is fully factually consistent.'
    ] * 2
    short_assessment_result = 'Fully Consistent'
```

tests/metrics/eval_clients/_openai.py
```diff
@@ -0,0 +1,111 @@
...
+                                        unstructured_assessment_result,
+                                        score_map)
+        assert len(scores) == len(unstructured_assessment_result)
+        for score in scores:
+            assert score == 1.0
+
+
+def test_get_text_response_azure_openai():
+    prompts = ['What is the capital of France?'] * 2
+    answer = 'Paris is the capital of France.'
```
yosukehigashi, line 63:
```suggestion
    prompts = ['Assess the factual consistency of the generated output...'] * 2
    answer = 'The output is fully factually consistent.'
```

tests/metrics/eval_clients/_openai.py
```diff
@@ -0,0 +1,111 @@
...
+        assert len(responses) == len(prompts)
+        for response in responses:
+            assert response == answer
+
+
+def test_get_float_score_azure_openai():
+    unstructured_assessment_result: list[str | None] = [
+        'Paris is the capital of France.'
+    ] * 2
+    short_assessment_result = 'Paris'
```
yosukehigashi, line 87:
```suggestion
    unstructured_assessment_result: list[str | None] = [
        'The output is fully factually consistent.'
    ] * 2
    short_assessment_result = 'Fully Consistent'
```

yosukehigashi (commented):
Reviewed the tests as well!

src/langcheck/metrics/de/reference_free_text_quality.py
```diff
@@ -316,172 +226,99 @@ def fluency(generated_outputs: List[str] | str,
...
+    fluency_template = get_template('de/metrics/fluency.j2')
 
     fluency_assessment_to_score = {
-        'Schlecht': 0,
-        'Ausreichend': 0.5,
-        'Gut': 1.0,
+        'Poor': 0,
+        'Fair': 0.5,
+        'Good': 1.0,
     }
```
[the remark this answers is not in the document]
liwii: No, I did that for consistency. All the Japanese prompts use English options, and because of the non-ascii characters I needed to replace some German options to English too. It is weird if the language used as the options changes based on whether the characters like "ä", "ö" are used or not, so I replaced all of them with English words.

src/langcheck/metrics/eval_clients/_openai.py
```diff
@@ -0,0 +1,352 @@
...
+            OpenAI), "Only sync clients are supported for similarity scoring."
+        return OpenAISimilarityScorer(openai_client=self._client,
+                                      openai_args=self._openai_args)
+
+
+class AzureOpenAIEvalClient(OpenAIEvalClient):
+
+    def __init__(self,
+                 text_model_name: str | None = None,
+                 embedding_model_name: str | None = None,
```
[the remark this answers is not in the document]
liwii: > make it clearer that the model name is required

That's one point, I also wanted to use the same EvalClient for text-based and embedding-based metrics. If you set both, you just need to pass the `EvalClient` to each metric and no need of caring if the metric is text-based or embedding-based anymore.

src/langcheck/metrics/en/pairwise_text_quality.py
```diff
@@ -61,203 +45,64 @@ def pairwise_comparison(
...
-            default 'openai'
-        openai_client: OpenAI or AzureOpenAI client, default None. If this is
-            None, we will attempt to create a default client.
-        openai_args: Dict of additional args to pass in to the
-            ``client.chat.completions.create`` function, default None
-        use_async: Whether to use the asynchronous API of OpenAI, default False
+        eval_model: The EvalClient instance used for the evaluation. This is
+            marked as Optional because of the backwards compatibility with the
+            previous order of the arguments, but you need to pass this argument
+            to make the function work.
```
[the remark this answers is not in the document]
liwii: Hmm, I don't fully agree because it is not just for backwards compatibility but also for consistency with other metrics. For the metrics that offers `'local'` option, I think it is the right design to pass `'local'` as the default option. If we keep this for metrics with 'local', even if users try to use all the metrics with OpenAI, they need to pass the `EvalClient` in a different position of the arguments, which doesn't make much sense to me.

Making this a kwarg also opens the possibility of adding `local` models for some metrics without further breaking the backward compatibility in the future.

src/langcheck/metrics/prompts/en/metrics/pairwise_comparison.j2
liwii: Totally makes sense. Updated the prompts.

liwii:
@yosukehigashi 

Mostly resolved your comments. I left a few still open for discussion. Also replied to your possibly related comment in #111. https://github.com/citadel-ai/langcheck/pull/111#issuecomment-2076537935

src/langcheck/metrics/en/pairwise_text_quality.py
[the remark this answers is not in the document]
yosukehigashi: Ah yeah that's true, if we add a `local` option it'll make sense to keep this `eval_model` arg in its current place. Maybe we can update the docstring here to reflect that? E.g. "This is marked as Optional so that it can follow the above arguments that have default values (for consistency with the other metrics), but this is in fact a required argument."

(Another option would be to make it like this, but it's probably a bit overkill 😅)
```
def pairwise_comparison(
        generated_outputs_a: List[str] | str,
        generated_outputs_b: List[str] | str,
        prompts: List[str] | str,
        sources_a: Optional[List[str] | str] = None,
        sources_b: Optional[List[str] | str] = None,
        reference_outputs: Optional[List[str] | str] = None,
        enforce_consistency: bool = True,
        *,
        eval_model: EvalClient) -> MetricValue[Optional[float]]:
```

src/langcheck/metrics/eval_clients/_openai.py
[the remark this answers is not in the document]
yosukehigashi: > I also wanted to use the same EvalClient for text-based and embedding-based metrics.

ohh I see, makes sense!

src/langcheck/metrics/de/reference_free_text_quality.py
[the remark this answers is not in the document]
yosukehigashi: Ok sgtm! Yeah I changed the Japanese options to English ones because OpenAI sometimes went haywire. As long as the descriptions of the options remain in the same language it seems like it works fine

yosukehigashi (approved):
Approving, since the last remaining comment is just about a docstring!

src/langcheck/metrics/en/pairwise_text_quality.py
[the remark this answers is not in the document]
liwii: SGTM, updated the docstrings following your idea!!

(Yeah I think making this keyword-only argument is a bit too much... let me go with the docstring update).

PR #111, opened 2024-04-19
Prototype of Claude-based evaluation

liwii:
Added the support of anthropic models as the evaluator.

To enable this, you need to install `langcheck[anthropic]` or `langcheck[all]`.

Example usage with dotenv:

```python

import langcheck
from langcheck.metrics.eval_clients import AnthropicEvalClient

# Set ANTHROPIC_API_KEY in .env
load_dotenv()

# The default model is Haiku
client = AnthropicEvalClient()

evals_toxicity = langcheck.metrics.toxicity(
    ['I am feeling pretty good!!!', 'F**k you!!'], eval_model=client)
 ```

### CODE REVIEW

**Comment by liwii** [comment]
@yosukehigashi @kennysong 

Here is the prototype of the Claude-based evaluation. Here is the example usage:

```python

import langcheck
from langcheck.metrics.eval_clients import AnthropicEvalClient

# Set ANTHROPIC_API_KEY in .env
load_dotenv()

# The default model is Haiku
client = AnthropicEvalClient()

evals_toxicity = langcheck.metrics.toxicity(
    ['I am feeling pretty good!!!', 'F**k you!!'], eval_model=client)
 ```

File: pyproject.toml [hunk 1]
```diff
@@ -58,6 +58,12 @@ all = [
     'langcheck[ja-optional]',
     'langcheck[zh]'
 ]
+
+# Different LLM clients
+anthropic = [
+    'anthropic'
+]
```
yosukehigashi, line 65: I feel like it'll be easier to just include this in the core deps for now, WDYT?

src/langcheck/metrics/eval_clients/__init__.py
```diff
@@ -1,8 +1,10 @@
+from langcheck.metrics.eval_clients._anthropic import AnthropicEvalClient
```
yosukehigashi, line 1: (Related to comment on deps) If we don't put `anthropic` in the core deps, we wrap this in a try/except

src/langcheck/metrics/eval_clients/_anthropic.py
```diff
@@ -0,0 +1,185 @@
...
+    '''EvalClient defined for Anthropic API.
+    '''
+
+    def __init__(self,
+                 anthropic_client: Anthropic | None = None,
+                 anthropic_args: dict[str, Any] | None = None,
+                 *,
+                 use_async: bool = False):
+        '''
+        Intialize the Anthropic evaluation client. The asuthentication
```
yosukehigashi, line 24:
```suggestion
        Initialize the Anthropic evaluation client. The authentication
```

src/langcheck/metrics/eval_clients/_anthropic.py
```diff
@@ -0,0 +1,185 @@
...
+                _call_api_with_exception_filter(model_input)
+                for model_input in tqdm_wrapper(model_inputs,
+                                                desc=tqdm_description)
+            ]
+
+        # Filter out exceptions and print them out.
+        for i, response in enumerate(responses):
+            if not isinstance(response, Exception):
+                continue
+            print('OpenAI failed to return an assessment corresponding to '
```
yosukehigashi, line 88:
```suggestion
            print('Anthropic failed to return an assessment corresponding to '
```

src/langcheck/metrics/eval_clients/_anthropic.py
```diff
@@ -0,0 +1,185 @@
...
+        return responses
+
+    def get_text_responses(
+            self,
+            prompts: Iterable[str],
+            *,
+            tqdm_description: str | None = None) -> list[str | None]:
+        '''The function that gets resonses to the given prompt texts.
+        We use Anthropic's 'claude-3-haiku-20240307' model by default, but you
+        can configure it by passing the 'model' parameter in the openai_args.
```
yosukehigashi, line 100:
```suggestion
        can configure it by passing the 'model' parameter in the anthropic_args.
```

src/langcheck/metrics/eval_clients/_anthropic.py
```diff
@@ -0,0 +1,185 @@
...
+        tqdm_description = tqdm_description or 'Scores (2/2)'
+        responses = self._call_api(prompts=get_score_prompts,
+                                   config=config,
+                                   tqdm_description=tqdm_description)
+        raw_response_texts = [
+            response.content[0].text if response else None
+            for response in responses
+        ]
+
+        def _turn_to_score(response: str | None) -> float | None:
```
yosukehigashi, line 171: I think we can use function calling in Anthropic as well:
- https://docs.anthropic.com/claude/docs/tool-use
- https://docs.anthropic.com/claude/docs/tool-use-examples#json-mode

yosukehigashi (commented):
It's great that adding a new evaluator is so simple now!

This might be more related to https://github.com/citadel-ai/langcheck/pull/110, but I feel like we want to keep the possibility open for using different prompts for different eval clients. I think it's still quite common for people to use slightly different prompting strategies for different models (e.g. OpenAI vs. Anthropic).

Instead of calling `get_template` in each of the metric functions, maybe we can change it so that the metric function passes some vars to the EvalClient, which in turn calls `get_template(f'{lang}/metrics/anthropic/{metric_name}.j2')`? WDYT?

src/langcheck/metrics/eval_clients/_anthropic.py
liwii: As it is mentioned in the API document, the feature is still in public beta, and it is not supported by the `anthropic` Python package. Practically even with Haiku this method seems to work quite well, so I would like to go with this for now. WDYT?

liwii:
I agree that the users might want to use different prompts for different LLM types, but I still think it is enough to store one set of prompts for each metric x language with the following reasons:

1. I don't think the current prompts are "optimized for OpenAI models"
1. There would be different prompting strategies for different models within each company too. It doesn't make sense to me that we only natively support `openai/` or `anthropic/` as the default set of the prompts.
1. If we need to rewrite the whole set of prompts for each `EvalClient` type, the implementation cost would become so much.
1. Even if we can create a set of tailor-made prompts for each `EvalClient`, we don't have good benchmarks to measure whether the evaluation quality is actually improved. 
1. One of the main points of this `EvalClient` design is that if you follow the interface you can easily implement a new `EvalClient` with newly published models (e.g. Llama 3). If `EvalClient` is heavily tied with the structure of `prompts/` directory of langcheck, it would become much harder.

I think we would rather allow users to add their custom prompts more easily in some way. I haven't come up with the good interface, and I'm not sure whether we should do that in #110 yet. WDYT?

yosukehigashi:
I agree with most of your points, so keeping this current structure sgtm. But two points I disagree with:
> There would be different prompting strategies for different models within each company too. It doesn't make sense to me that we only natively support openai/ or anthropic/ as the default set of the prompts.

Agreed that different people will have different preferences for the behavior of an evaluator (and so would have different prompting strategies), but IMO that's a separate issue. If Prompt A works way better _in general_ for OpenAI, and likewise Prompt B for Anthropic, then we should for sure be using Prompt A with OpenAI and Prompt B for Anthropic. So if such a Prompt A and Prompt B existed, I'd argue that it makes sense to just natively support openai/ and anthropic/.

> One of the main points of this EvalClient design is that if you follow the interface you can easily implement a new EvalClient with newly published models (e.g. Llama 3). If EvalClient is heavily tied with the structure of prompts/ directory of langcheck, it would become much harder.

If the prompts are unchangeable for all EvalClients, then you'll be out of luck if you want to introduce a new EvalClient (e.g. Llama3) and for whatever reason it performs poorly with the current prompt. Also, if all EvalClients use the same prompt, then it becomes harder to change the _prompt_ as well, since it might be a good change for some EvalClients but not good for others.

That said, I definitely agree with (1) and (4), meaning that I wouldn't really know what to do differently for the Anthropic prompts and would probably just end up copy-pasting everything 😅 The only thing that I know of is that [Anthropic likes XML tags](https://docs.anthropic.com/claude/docs/use-xml-tags) whereas OpenAI tends to like Markdown format, but that just seems kind of minor to warrant a rewrite of all the prompts..

However, if in the future we _do_ find that different prompting strategies work much better for certain EvalClients, IMO we should support that natively somehow. I don't think it's a good UX if in `ImaginaryEvalClient`, we have a docstring that's like `Note: ImaginaryEvalClient's factual_consistency metric performs better when you when you prompt it like this: <some prompt>. Change the prompt to this by passing it in as the "custom_prompt" for optimal performance.`

src/langcheck/metrics/eval_clients/_anthropic.py
yosukehigashi: > it is not supported by the anthropic Python package

Wait really?😨  Where do you see that?

But yeah it might be better to wait for it to be more widely released, esp. considering this line: "Tool use is not yet available on third-party platforms like Vertex AI or AWS Bedrock, but is coming soon". We can keep your approach for now.

I've also been meaning to look into using https://github.com/jxnl/instructor, and this might be a good opporunity since they seem to have support for Anthropic (caveat: I haven't checked the implementation, so maybe they do something similar to your approach anyway).

liwii: Here is the official `anthropic-tools` package:
https://github.com/anthropics/anthropic-tools

It says:
> This SDK is Currently in Alpha and meant ONLY AS A RESEARCH PREVIEW. We promise no ongoing support. It is not intended for production use. Production-ready function calling support is coming to the Anthropic API soon.

liwii: I looked into `instructor`'s implementation, and it seems pretty useful!!

I overlooked that, but `anthropic-sdk-python`'s [README](https://github.com/anthropics/anthropic-sdk-python?tab=readme-ov-file#tool-use-beta) says
> This SDK provides beta support for tool use, aka function calling. More details can be found in [the documentation](https://docs.anthropic.com/claude/docs/tool-use).

and `instructor` uses this when `mode=ANTHROPIC_TOOLS` parameter is given.

Sorry for not checking that carefully, but now I'm okay with either
- using the `tools` API
- using `instructor`
- keeping the current structure

Do you have any preferences?

yosukehigashi: > Here is the official `anthropic-tools` package:
> https://github.com/anthropics/anthropic-tools

Hmm I think this is an older repo or something? It looks like the Beta is in the normal `anthropic` package: https://github.com/anthropics/anthropic-sdk-python?tab=readme-ov-file#tool-use-beta

liwii: sgtm!!

liwii:
Your statements seem valid to me, but I'm still not sure whether the premises hold:

>  If Prompt A works way better in general for OpenAI, and likewise Prompt B for Anthropic
> if you want to introduce a new EvalClient (e.g. Llama3) and for whatever reason it performs poorly with the current prompt.


Given the performance and the improvement speed of the latest models, I cannot imagine a situation that
- Some models can potentially perform pretty well
- but they perform poorly with the certain prompts

that might happen to some low-tier models (e.g. gpt-3.5 could perform significantly better with English prompts compared to Japanese prompts), but in that case vertical splits (e.g. gpt-4 vs gpt-3.5) makes more sense than horizontal split (e.g. OpenAI vs Anthropic).

I totally agree that we might want to support some prompts natively, but in terms of the code structure, I would like to start thinking after getting a concrete problem.

yosukehigashi:
Yeah I don't think the more recent models will suddenly start outputting gibberish due to a slight prompt change or anything like that, but I can definitely see us running into a situation where a prompt change causes the performance of one EvalClient to improve but another to regress (once we can measure these kinds of things rigorously).

> I totally agree that we might want to support some prompts natively, but in terms of the code structure, I would like to start thinking after getting a concrete problem.

Yep sgtm!

yosukehigashi:
I've been playing around with it a bit and it seems to work pretty well! We should get this in soon and start testing it more.

It might be good to turn down the temperature to 0 to make the results a bit more deterministic (https://docs.anthropic.com/claude/docs/troubleshooting#common-issues--mitigation-strategies-with-llms)
image

pyproject.toml
liwii: I agree it'll be easier, but when I looked into different LLM-based tools with multiple platform types, tools like `llama-index` or `langchain`offer different subpackages for non-OpenAI clients, and `instructor` mark the non-OpenAI dependencies as optional.

It makes sense because most of the users will only use one or two client types, so it is not good to install all the other unused API sdks by default. We need to do this for `gemini` anyways because of the Python version issue.

Therefore, though it is not a strong opinion, I would prefer the current optional dependency style. WDYT?

src/langcheck/metrics/eval_clients/_anthropic.py
```diff
@@ -0,0 +1,189 @@
...
+        Args:
+            prompts: The prompts you want to get the responses for.
+
+        Returns:
+            A list of responses to the prompts. The responses can be None if the
+            evaluation fails.
+        '''
+        config = {
+            "model": "claude-3-haiku-20240307",
+            "max_tokens": 1024,
```
yosukehigashi, line 111: nit: I feel like it's fine to keep this at the default value?

tests/metrics/eval_clients/_anthropic.py
```diff
@@ -0,0 +1,54 @@
...
+
+import os
+from unittest.mock import Mock, patch
+
+from anthropic.types.message import Message
+
+from langcheck.metrics.eval_clients import AnthropicEvalClient
+
+
+def test_get_text_response_openai():
```
yosukehigashi, line 11: Anthropic

tests/metrics/eval_clients/_anthropic.py
```diff
@@ -0,0 +1,54 @@
...
+        # Set the necessary env vars for the AnthropicEvalClient
+        os.environ["ANTHROPIC_API_KEY"] = "dummy_key"
+        client = AnthropicEvalClient()
+        responses = client.get_text_responses(prompts)
+        assert len(responses) == len(prompts)
+        for response in responses:
+            assert response == answer
+
+
+def test_get_float_score_openai():
```
yosukehigashi, line 30: Anthropic

yosukehigashi (commented):
Code LGTM! I'll test things out a bit more once I figure out how to create a company account

src/langcheck/metrics/eval_clients/_anthropic.py
liwii: Actually you need to specify a value, otherwise you get the following error

> Anthropic failed to return an assessment corresponding to 0th prompt: Missing required arguments; Expected either ('max_tokens', 'messages' and 'model') or ('max_tokens', 'messages', 'model' and 'stream') arguments to be given


I updated the limit to 4096

yosukehigashi (approved):
LGTM!🎉

PR #112, opened 2024-05-03
Update devcontainer.json

liwii:
The `postCreateCommand` of our default VSCode dev container was failing because the container was missing `cargo`, which is required by some libraries in `langcheck[ja-optional]` dependencies. Fixed this problem and re-formatted the `devcontainer.json` file.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -65,13 +65,20 @@ In this case, you need have a Rust compiler in your environment so that the pack
 
 Follow the [official documentation](https://www.rust-lang.org/tools/install) and install Rust before installing `langcheck`.
 
-### 3. The error `AttributeError: module 'langcheck.metrics' has no attribute 'ja'`
+If you do not need the `[ja-optional]` dependencies, you can also install each language individually by
+
+```sh
+pip install langcheck[en,ja,de,zh]
+```
+
+
+### 4. The error `AttributeError: module 'langcheck.metrics' has no attribute 'ja'`
 
 If you see this error when calling a LangCheck function, such as `langcheck.metrics.ja.toxicity()`, you probably need to install the Japanese LangCheck package.
 
 Run `pip install langcheck[ja]` (or for your required language) and try again.
 
-### 4. The error `ModuleNotFoundError`
+### 5. The error `ModuleNotFoundError`
 
 If you see this error when importing a LangCheck package, such as `import langcheck.metrics.ja`, you probably need to install the Japanese LangCheck package.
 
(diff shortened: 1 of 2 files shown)


kennysong:
Hmm I thought we fixed the rust compiler issue in https://github.com/citadel-ai/langcheck/pull/45.

Do you know which dependency this is breaking on? 

I guess our install tests didn't catch it since we only test on Mac ARM64/Linux x86/Windows x86, not Linux ARM64. (Just signed up for [the ARM GitHub Runners beta](https://resources.github.com/devops/accelerate-your-cicd-with-arm-and-gpu-runners-in-github-actions/))

kennysong:
Probably worth adding an entry for this in the FAQ after the gcc question: https://langcheck.readthedocs.io/en/latest/installation.html#the-error-message-command-gcc-failed

liwii:
Yeah that was `sudachipy`. I checked the latest distributions of the package. and there is no wheel for ARM64 Linux.

https://pypi.org/project/SudachiPy/#files

Let me add an entry to the FAQ about it.

liwii:
Updated `instruction.md`. Please double-check!!

docs/installation.md
```diff
@@ -58,6 +58,13 @@ Most systems already have `gcc` installed, with the exception of "slim" Docker i
...
 1. Run `apt update && apt install build-essential -y` (e.g. in your Dockerfile) to install gcc.
 
+### 3. The error message `Failed to build sudachipy`
+
+The `sudachipy` package, which is installed with `langcheck[all]` or `langcheck[ja-optional]`, does not have a pre-built wheel for ARM64 Linux devices.
+In this case, you need have a Rust compiler in your environment so that the package can be built in the local environment.
+
+Follow the [official documentation](https://www.rust-lang.org/tools/install) and install Rust before installing `langcheck`.
+
 ### 3. The error `AttributeError: module 'langcheck.metrics' has no attribute 'ja'`
```
yosukehigashi, line 68: Please update the numbers, here and below

docs/installation.md
```diff
@@ -58,6 +58,13 @@ Most systems already have `gcc` installed, with the exception of "slim" Docker i
 1. Use `python:3.11-buster` instead of `python:3.11-slim-buster`, which includes gcc.
 1. Run `apt update && apt install build-essential -y` (e.g. in your Dockerfile) to install gcc.
 
+### 3. The error message `Failed to build sudachipy`
+
+The `sudachipy` package, which is installed with `langcheck[all]` or `langcheck[ja-optional]`, does not have a pre-built wheel for ARM64 Linux devices.
```
yosukehigashi, line 63: Maybe we should also suggest installing each lang individually, i.e. `langcheck[en,ja,de,zh]`?

liwii: Added the lines

yosukehigashi (approved):
LGTM!

PR #113, opened 2024-05-06
Add Gemini Eval Client

yosukehigashi:
Adds the Gemini Eval Client. It mostly follows https://github.com/citadel-ai/langcheck/pull/111, but with a few notable differences:
- It does not yet support `use_async`, since there's some issue in the Gemini API (see TODO in code)
- Support for `similarity_scorer` is added
- It leverages the function calling API
- Gemini requires Python 3.9+, so I updated the devcontainer to use 3.9. Note that LangCheck itself is still compatible with 3.8 if you don't install `langcheck[gemini]`

There are three ways to intantiate a GeminiEvalClient
```
import os
import google.generativeai as genai

from langcheck.metrics.eval_clients import GeminiEvalClient

os.environ["GOOGLE_API_KEY"] = 'YOUR_GOOGLE_API_KEY'

# Default client (uses gemini-pro)
client = GeminiEvalClient()

# Client from model
model = genai.GenerativeModel('models/gemini-1.5-pro-latest')
client = GeminiEvalClient(model=model)

# Client from model args
args = {'model_name': 'gemini-1.5-pro-latest'}
client = GeminiEvalClient(model_args=args)
```


src/langcheck/metrics/eval_clients/_gemini.py
```diff
@@ -0,0 +1,242 @@
...
+        Returns:
+            A list of scores for the given prompts. The scores can be None if
+            the evaluation fails.
+        '''
+        if language not in ['en', 'ja', 'de', 'zh']:
+            raise ValueError(f'Unsupported language: {language}')
+
+        # Gemini's function calling API is similar to OpenAI's, so we can reuse
+        # the prompt template.
+        fn_call_template = get_template(f'{language}/get_score/openai.j2')
```
liwii, line 163: I think it is better to rename the prompts to `function_calling.j2` and `plain_text.j2` now.. What do you think?

src/langcheck/metrics/eval_clients/_gemini.py
```diff
@@ -0,0 +1,242 @@
...
+                  tqdm_description: str | None = None) -> list[Any]:
+        # A helper function to call the API with exception filter for alignment
+        # of exception handling with the async version.
+        def _call_api_with_exception_filter(prompt: str) -> Any:
+            try:
+                return self._model.generate_content(prompt, **config)
+            except Exception as e:
+                return e
+
+        if self._use_async:
```
liwii, line 72: If the branch is not used, I think it is less confusing to comment the lines out or just remove them.

pyproject.toml
```diff
@@ -56,14 +56,18 @@ zh = [
...
 
 all = [
     'langcheck[de]',
     'langcheck[en]',
     'langcheck[ja]',
     'langcheck[ja-optional]',
     'langcheck[zh]',
-    'langcheck[anthropic]'
+    'langcheck[anthropic]',
+    'langcheck[gemini]'
```
liwii, line 70: Maybe it is better to use `[google]` here following `[anthropic]`?

src/langcheck/metrics/eval_clients/_gemini.py
```diff
@@ -0,0 +1,242 @@
...
+                for the given assessment prompts.
+            score_map: The mapping from the short assessment results
+                (e.g. "Good") to the scores.
+            tqdm_description: The description to be shown in the tqdm bar.
+
+        Returns:
+            A list of scores for the given prompts. The scores can be None if
+            the evaluation fails.
+        '''
+        if language not in ['en', 'ja', 'de', 'zh']:
```
liwii, line 158: There is no prompts for `zh` (it reuses the English prompts). In that sense you should remove it from the list.

liwii (commented):
Basically LGTM, left a few minor comments

src/langcheck/metrics/eval_clients/_gemini.py
yosukehigashi: Yep good point! Renamed the files

pyproject.toml
yosukehigashi: hmm I feel like `gemini` is more consistent actually, since we call the eval client `GeminiEvalClient` and not `GoogleEvalClient` - wdyt?

src/langcheck/metrics/eval_clients/_gemini.py
yosukehigashi: Removed!

pyproject.toml
liwii: Makes sense!!

liwii (approved):

PR #116, opened 2024-05-08
Rename ja & de prompts

liwii:
In https://github.com/citadel-ai/langcheck/pull/113 the prompt files to extract scores from the LLM outputs are renamed. However, for ja & de language the prompt names remained the same so all `ja` & `de` metrics with `EvalClient` does not work now.

Renamed the prompts and confirmed that they work with `AzureOpenAIEvalClient`, `AnthropicEvalClient` and `GeminiEvalClient`.


kennysong:
LGTM. Why didn't tests catch this?

kennysong (approved):

liwii:
Because the tests always use `en` as the language parameter.

I think we should parameterize the tests with the available languages. Let me add that in this PR too.

liwii:
Parameterized the test cases.

I also found that the tests were not running because the filenames didn't start with `test_`... We didn't notice that because we were running individual tests in the local environment.

Already fixed and they should run in the Github actions now.

kennysong (approved):
LGTM if the tests pass!

yosukehigashi (approved):
Thanks for the fix and for improving the tests!🙇

PR #119, opened 2024-05-13
Bump version to 0.8.0.dev1

yosukehigashi:
Bumps the version to a dev version. This is a pre-release, so pip will ignore it by default (https://pip.pypa.io/en/stable/cli/pip_install/#pre-release-versions).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "langcheck"
-version = "0.7.1"
+version = "0.8.0.dev1"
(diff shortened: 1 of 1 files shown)


kennysong:
LGTM! Do you want to publish a release instead of installing from a tag though?

https://stackoverflow.com/questions/13685920/install-specific-git-commit-with-pip

kennysong (approved):

yosukehigashi:
> LGTM! Do you want to publish a release instead of installing from a tag though?
> 
> https://stackoverflow.com/questions/13685920/install-specific-git-commit-with-pip

Resolved offline!

PR #120, opened 2024-05-28
Custom evaluator metric

yosukehigashi:
Adds a `custom_evaluator` metric that allows users to specify a custom prompt-based metric.

E.g. you can create your own "Answer Conciseness" metric by first writing a prompt template in a jinja file:
```
You are evaluating the conciseness of the answer to a user's query. Here is the data:
[BEGIN DATA]
************
[User Query]: {{ user_query }}
************
[Answer]: {{ gen_output }}
************
[END DATA]

Determine whether the answer is a concise response to the user's query.
The available assessments are:
`Concise` - The answer is a concise response to the user's query that is short and to the point.
`Not Concise` - The answer is not a concise response to the user's query, e.g. there are unnecessary details or the answer is too long.

Write out a one-sentence justification before giving the final assessment.
```

And then, you can compute the metric like this:
```

gen_output = [
    '4',
    'The answer to the question 2+2 is 4. If you need any more assistance with math, feel free to ask!'
]
prompt = ['what is 2+2?', 'what is 2+2?']

score_map = {'Not Concise': 0.0, 'Concise': 1.0}

langcheck.metrics.custom_evaluator(generated_outputs=gen_output,
                                   prompts=prompt,
                                   sources=None,
                                   reference_outputs=None,
                                   eval_model=eval_client,
                                   metric_name='answer_conciseness',
                                   score_map=score_map,
                                   template_path='answer_conciseness.j2',
                                   language='en')
```
image

### DIFF
--- a/src/langcheck/metrics/custom_text_quality.py
+++ b/src/langcheck/metrics/custom_text_quality.py
@@ -20,6 +20,22 @@ def custom_evaluator(generated_outputs: list[str] | str | None,
     assess the provided inputs using the prompt template, and then convert those
     assessments into scores using the score map.
 
+    The prompt template should be a Jinja2 file (file extension .j2) that
+    specifies the criteria that an LLM (as configured in the Eval Client) should
+    follow when evaluating an instance. The template is allowed to have
+    placeholders for the following variables (NOTE: not all are required):
+    - `gen_output`: The generated output
+    - `user_query`: The prompt
+    - `src`: The source text
+    - `ref_output`: The reference output
+
+    The prompt template should also specify the final available assessments for
+    the LLM evaluator, e.g. "Good", "Bad", "Neutral", etc. The score map should
+    then map each of those available assessments to a numerical score. E.g. if
+    the available assessments in the prompt template are "Good", "Bad", and
+    "Neutral", the score map should be something like:
+    ``score_map = {'Good': 1.0, 'Neutral': 0.5, 'Bad': 0.0}``
+
     Args:
         generated_outputs: The model generated output(s)
         prompts: The prompts used to generate the output(s)
@@ -28,7 +44,8 @@ def custom_evaluator(generated_outputs: list[str] | str | None,
         eval_model: The EvalClient instance used for the evaluation
         metric_name: The name of the metric
         score_map: A dictionary mapping the evaluator's assessments to scores
-        template_path: The path to the prompt template file
+        template_path: The path to the prompt template file. This should be a
+            Jinja2 file (file extension .j2).
         language: The language that the evaluator will use ('en', 'ja', or 'de')
 
     Returns:
--- a/src/langcheck/metrics/custom_text_quality.py
+++ b/src/langcheck/metrics/custom_text_quality.py
@@ -36,6 +36,12 @@ def custom_evaluator(generated_outputs: list[str] | str | None,
     "Neutral", the score map should be something like:
     ``score_map = {'Good': 1.0, 'Neutral': 0.5, 'Bad': 0.0}``
 
+    NOTE: We have found that LLM models sometimes behave weirdly when the
+    assessments are non-ascii characters (see
+    https://github.com/citadel-ai/langcheck/pull/84 as an example). So, we
+    recommend making the final assessments ascii characters, even when the rest
+    of the prompt template contains non-ascii characters (e.g. Japanese).
+
     Args:
         generated_outputs: The model generated output(s)
         prompts: The prompts used to generate the output(s)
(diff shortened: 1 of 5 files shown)


src/langcheck/metrics/custom_text_quality.py
```diff
@@ -0,0 +1,104 @@
...
+def custom_evaluator(generated_outputs: list[str] | str | None,
+                     prompts: list[str] | str | None,
+                     sources: list[str] | str | None,
+                     reference_outputs: list[str] | str | None,
+                     eval_model: EvalClient, metric_name: str,
+                     score_map: dict[str, float], template_path: str,
+                     language: str) -> MetricValue[float | None]:
+    '''Calculates the scores of a custom evaluator. The EvalClient will first
+    assess the provided inputs using the prompt template, and then convert those
+    assessments into scores using the score map.
```
liwii, line 21: We need to explain what parameters are allowed in Jinjia templates here (probably want a `.md` page for that too)

src/langcheck/metrics/custom_text_quality.py
```diff
@@ -0,0 +1,104 @@
...
+        language: The language that the evaluator will use ('en', 'ja', or 'de')
+
+    Returns:
+        A MetricValue object
+    '''
+    generated_outputs, prompts, reference_outputs, sources = validate_parameters_custom_evaluator(  # NOQA: E501
+        generated_outputs, prompts, reference_outputs, sources)
+    # Find the length of the first non-None list (they are guaranteed to all be
+    # the same length)
+    num_examples = next(
```
liwii, line 41: No a strong opinion, but maybe we can do that in the `validate_parameters_` function?

src/langcheck/metrics/custom_text_quality.py
liwii: Besides the allowed parameters, we also mention some interface the EvalClient expects such as
- The LLM needs to choose one "Evaluation Result" option from the ones listed in the prompt
- These options should correspond to the keys of `score_map`
- The keys should not contain non-ascii characters

src/langcheck/metrics/custom_text_quality.py
```diff
@@ -0,0 +1,104 @@
...
+        f'The prompt template contains invalid parameters. The allowed parameters are {allowed_params} but the prompt template expects the parameters {expected_params}'  # NOQA: E501
+    expected_param_to_arg = {
+        'gen_output': generated_outputs,
+        'user_query': prompts,
+        'src': sources,
+        'ref_output': reference_outputs,
+    }
+    for param in expected_params:
+        assert expected_param_to_arg[param] is not None, \
+            f'The prompt template expects the parameter "{param}" but it is not provided.'  # NOQA: E501
```
liwii, line 69: You might want to add `{% if src is none... %}` condition as we do for pairwise comparison, maybe it is better to remove the validation?

liwii (commented):
Nice work! I tested with my custom japanese evaluation metric, and it worked pretty well!!

Left a few comments mainly about documentation & validation

src/langcheck/metrics/custom_text_quality.py
yosukehigashi: You mean that we should return `num_examples` from `validate_parameters_custom_evaluator`?

src/langcheck/metrics/custom_text_quality.py
yosukehigashi: That's a fair point, but I still think it's worth validating this, since the behavior can be surprising otherwise. E.g. if the user has `{% if src is none... %}` but then passes in `sources=None`, the individual `src` vars actually become empty strings and not `None`.

So if the user has logic such as `{% if src is none... %}` in their prompt template, IMO it's better for them to pass in a list of Nones as `sources`, WDYT?

src/langcheck/metrics/custom_text_quality.py
yosukehigashi: Added this info in the function docstring! Agreed that we'd want to add a page in readthedocs too - I can add that in a follow-up PR

src/langcheck/metrics/custom_text_quality.py
liwii: Yes!
But now I think it could mess up the return value type, let's leave it as it is

src/langcheck/metrics/custom_text_quality.py
liwii: Makes sense!

liwii (approved):
Sorry for being late, let me approve it!!

PR #124, opened 2024-06-15
Return `None` if the function calling step returns an invalid assessment

yosukehigashi:
Updates the eval clients to set a metric value to `None` in the rare case that the function calling step returns an invalid assessment. Otherwise, the evaluator will fail with a KeyError.

(I tried for a bit to trigger a bad assessment but couldn't get it to happen 😅)

Also bump the version to `0.8.0.dev3`


liwii (approved):
Nice!!

PR #125, opened 2024-06-21
[WIP] Introduce Ruff as the formatter

liwii:


liwii:
Made a draft PR to apply Ruff as the formatter for trial. I'm not thinking of merging it immediately, but take a look and tell me your impression when you have time!!

@yosukehigashi @kennysong

.github/workflows/formatting.yml
```diff
@@ -31,17 +31,9 @@ jobs:
...
-      # Run the flake8 format test
-      - name: Flake8
-        run: flake8 .
-
-      # Run the isort format test
-      - name: Isort
-        run: isort --check-only .
+      # Run the ruff format test
+      - name: Ruff
+        run: ruff check .
```
kennysong, line 36: We use `ruff check src/` in the Python client

pyproject.toml
```diff
@@ -87,20 +85,13 @@ dev = [
...
 
-[tool.yapf]
-based_on_style = "google"
-column_limit = 80
-
-[tool.yapfignore]
-ignore_patterns = [
-    "venv/*",
+[tool.ruff]
+lint.select = ["E4", "E7", "E9", "F", "I"]
```
kennysong, line 89: We also include "Q" in the Python client

pyproject.toml
```diff
@@ -87,20 +85,13 @@ dev = [
...
+line-length=80
```
kennysong, line 90: Is this not the default?

kennysong (commented):
Nice! I compared the configuration to the Lens client and browsed through the code changes. Looks pretty reasonable to me.

.github/workflows/formatting.yml
liwii: It was just because there were not unit tests for the Python client when we introduced ruff. But I think it might be better to specify the directories explicitly so I made a change like that!!

pyproject.toml
liwii: Good catch, added!!

liwii:
Actually this change could make a lot of merge conflicts when a new change is made. Let me mark it as ready and try to merge while there is no new PR!!

kennysong:
Sounds good! Let's make another PR in the Lens client to make sure the configs match. Let's merge this PR now.

kennysong (approved):

PR #129, opened 2024-07-08
Custom pairwise evaluator metric

yosukehigashi:
Adds a `custom_pairwise_evaluator` metric that allows users to specify a custom prompt-based pairwise metric. Similar to the pairwise comparison metric, `enforce_consistency` is enabled by default.

Some examples of metrics that you can create with this are:
- Pairwise answer correctness - which response (A or B) is more correct / closer to the reference answer?
- Pairwise toxicity - which response (A or B) is less toxic?
- Pairwise context relevance - which source text (A or B) is more relevant for answering the user's query?

E.g. you can create your own "Pairwise Answer Conciseness" metric with a prompt template like:
```
You are evaluating the conciseness of the answer to a user's query. Here is the data:
[BEGIN DATA]
************
[User Query]: {{ user_query }}
************
[Answer A]: {{ gen_output_a }}
************
[Answer B]: {{ gen_output_b }}
************
[END DATA]

Determine which of the responses is a more concise response to the user's query.
The available assessments are:
`Response A` - Response A is a more concise response.
`Response B` - Response B is a more concise response.
`Tie` - The two responses are roughly equal in conciseness.

Write out a one-sentence justification before giving the final assessment.
```

And then compute the metric like this:
```

gen_output_a = [
    "4",
    "The answer to the question 2+2 is 4. If you need any more assistance with math, feel free to ask!",
    "I cannot answer that question."
]
gen_output_b = [
    "The answer to the question 2+2 is 4. If you need any more assistance with math, feel free to ask!",
    "4",
    "I don't know the answer to that question."
]
prompt = ["what is 2+2?", "what is 2+2?", "what is 2+2?"]
score_map = {"Response A": 0.0, "Response B": 1.0, "Tie": 0.5}

langcheck.metrics.custom_pairwise_evaluator(
    generated_outputs_a=gen_output_a,
    generated_outputs_b=gen_output_b,
    prompts=prompt,
    sources_a=None,
    sources_b=None,
    reference_outputs=None,
    eval_model=eval_client,
    metric_name="answer_conciseness",
    score_map=score_map,
    template_path="pairwise_answer_conciseness.j2",
    language="en",
)
```
image

### DIFF
--- a/src/langcheck/metrics/_pairwise_text_quality_utils.py
+++ b/src/langcheck/metrics/_pairwise_text_quality_utils.py
@@ -81,6 +81,7 @@ def enforce_pairwise_comparison_consistency(
     original_explanations: list[str | None],
     swapped_scores: list[float | None],
     swapped_explanations: list[str | None],
+    score_map: dict[str, float],
 ) -> tuple[list[float | None], list[str | None]]:
     """Enforce consistency in pairwise comparison scores.
 
@@ -91,8 +92,18 @@ def enforce_pairwise_comparison_consistency(
         swapped_scores: The scores for the swapped order of the models
         swapped_explanations: The explanations for the swapped order of the
             models
+        score_map: The mapping from the assessment results to the scores (e.g.
+            {'Response A': 0.0, 'Response B': 1.0, 'Tie': 0.5})
     """
-    # Iterate through the scores and explanations to check for consistency.
+    # Get the sorted list of available scores
+    sorted_available_scores = list(score_map.values())
+    sorted_available_scores.sort()
+
+    # Iterate through the scores and explanations to check for consistency. A
+    # score is consistent if the score's index in the sorted_available_scores
+    # list is the inverse of the swapped score's index. For example, if the
+    # score_map is {'Response A': 0.0, 'Response B': 1.0, 'Tie': 0.5}, and the
+    # score is 0.0, the swapped score should be 1.0.
     # If a score is not consistent, set it to None, and merge the two
     # explanations to show the inconsistency.
     scores = original_scores.copy()
@@ -104,7 +115,12 @@ def enforce_pairwise_comparison_consistency(
             scores[i] = None
             explanations[i] = None
             continue
-        if scores[i] + swapped_scores[i] != 1.0:  # type: ignore
+
+        if (
+            sorted_available_scores.index(scores[i])  # type: ignore
+            + sorted_available_scores.index(swapped_scores[i])  # type: ignore
+            != len(sorted_available_scores) - 1
+        ):
             scores[i] = None
             explanations[i] = (
                 f"Original assessment: {explanations[i]}\nSwapped assessment: {swapped_explanations[i]}"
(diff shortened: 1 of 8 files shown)


src/langcheck/metrics/custom_text_quality.py
```diff
@@ -148,3 +154,232 @@ def _args_to_prompt_param(
...
+    - `src_b`: The source text for Model B
+    - `ref_output`: The reference output
+
+    The prompt template should also specify the final available assessments for
+    the LLM evaluator, e.g. "Response A", "Response B", "Tie", etc. The score
+    map should then map each of those available assessments to a numerical
+    score. E.g. if the available assessments in the prompt template are
+    "Response A", "Response B", and "Tie", the score map should be something
+    like:
+    ``score_map = {'Response A': 0.0, 'Response B': 1.0, 'Tie': 0.5}``
```
liwii, line 196: nit: it might be too obvious but we should mention that the options are symmetric between Response A and Response B, otherwise the consistency check will always fail.

src/langcheck/metrics/_pairwise_text_quality_utils.py
```diff
@@ -91,8 +92,18 @@ def enforce_pairwise_comparison_consistency(
...
+        score_map: The mapping from the assessment results to the scores (e.g.
+            {'Response A': 0.0, 'Response B': 1.0, 'Tie': 0.5})
     """
-    # Iterate through the scores and explanations to check for consistency.
+    # Get the sorted list of available scores
+    sorted_available_scores = list(score_map.values())
+    sorted_available_scores.sort()
+
+    # Iterate through the scores and explanations to check for consistency. A
+    # score is consistent if the score's index in the sorted_available_scores
```
liwii, line 103: nit: it would be clearer if you could move the sentence from `A score is consistent if...` to line 118. It seems a bit far from the actual logic.

liwii (approved):
Sorry for the late review! It worked well in my environment too so mostly LGTM, just added a few nit comments.

src/langcheck/metrics/_pairwise_text_quality_utils.py
yosukehigashi: makes sense, done!

src/langcheck/metrics/custom_text_quality.py
yosukehigashi: sorry for the dumb q, but what did you mean by symmetric here? or put another way, what would be an example of an asymmetric score map?

(Also noticed that I forgot to add the `enforce_consistency` docstring, so added that)

liwii: Hmm it might be too obvious, but for example, if there are 5 options, 1 (for Response A) should correspond to 5 (for Response B), 2 should correspond to 4, 3 should be completely in the middle.

I don't think there is any reasonable scoring that doesn't follow this, but since if you don't follow that rule `enforce_consistency` always fails, I think it's worth noting.

yosukehigashi: Ah ok I see what you mean. So if a user comes up with a scoring system like:
1 - Response A is way better
2 - Response A is slightly better
3 - Response B is slightly better

And Model A is always slightly better than Model B, then in the original order it'll always be 2 but then when swapped it'll always be 3.

I feel like it could be a bit distracting to put this in the top-level docstring though (I think the user would need to be intentionally unreasonable to come up with something like this 😆) so I'll add this as a code comment!

yosukehigashi: added a quick note in the docstring, and a long explanation in the code comments

liwii (approved):
LGTM!

PR #130, opened 2024-07-09
Answer correctness metric

yosukehigashi:
Implemented answer correctness for en and ja.

Here are some example test cases:

```

gen_output = [
    "Tokyo is the capital of Japan.",
    "Osaka is the capital of Japan.",
    "Washington, D.C. is the capital of the US.",
    "Many people consider New York City is the capital of the US. But it is technically Washington, D.C.",
    "The capital of BlahBlah land is Blahtropolis.",
    "I am not sure what the capital of BlahBlah land is.",
]
prompt = [
    "what is the capital of Japan?",
    "what is the capital of Japan?",
    "what is the capital of the US?",
    "what is the capital of the US?",
    "what is the capital of BlahBlah land?",
    "what is the capital of BlahBlah land?",
]
ref_outputs = [
    "Tokyo",
    "Tokyo",
    "Washington, D.C.",
    "Washington, D.C.",
    "Blahtropolis.",
    "Blahtropolis.",
]

langcheck.metrics.answer_correctness(
    gen_output, ref_outputs, prompt, eval_model=eval_client
)
```
image

```

gen_output = [
    "東京は日本の首都です。",
    "大阪は日本の首都です。",
    "ワシントンD.C.はアメリカの首都です。",
    "多くの人々はニューヨーク市がアメリカの首都だと考えています。しかし、実際はワシントンD.C.です。",
    "ブラブラランドの首都はブラトロポリスです。",
    "ブラブラランドの首都が何かわかりません。",
]
prompt = [
    "日本の首都は何ですか？",
    "日本の首都は何ですか？",
    "アメリカの首都は何ですか？",
    "アメリカの首都は何ですか？",
    "ブラブラランドの首都は何ですか？",
    "ブラブラランドの首都は何ですか？",
]
ref_outputs = [
    "東京",
    "東京",
    "ワシントンD.C.",
    "ワシントンD.C.",
    "ブラトロポリス。",
    "ブラトロポリス。",
]

langcheck.metrics.ja.answer_correctness(
    gen_output, ref_outputs, prompt, eval_model=eval_client
)
```
image


src/langcheck/metrics/en/reference_based_text_quality.py
```diff
@@ -4,14 +4,78 @@
...
+    Returns:
+        A :class:`~langcheck.metrics.metric_value.MetricValue` object
+    """
+    generated_outputs, reference_outputs, prompts = (
+        validate_parameters_answer_correctness(
+            generated_outputs, reference_outputs, prompts
+        )
+    )
+
+    answer_correctness_template = get_template(
```
liwii, line 48: For other English metrics, the templates are loaded with `load_prompt_template` function now to use specific prompts for certain models (only used for Prometheus for now)

Given there is no Prometheus prompt I think it is okay to call `get_template` here for now, but we need to think when to write these model-specific prompts in the future

src/langcheck/metrics/prompts/en/metrics/answer_correctness.j2
```diff
@@ -0,0 +1,18 @@
+You are evaluating the correctness of the answer to a user's query. Here is the data:
```
liwii, line 1: nit: even for humans it could be a bit confusing if the prompt has both "answer" and "reference answer"... Could we mention what this "reference answer" is here? (For Japanese too)

liwii (commented):
Mostly LGTM, one comment about the prompts and another comment about how to maintain the model-specific prompts in the future

src/langcheck/metrics/en/reference_based_text_quality.py
yosukehigashi: Oh nice! I wrote this before `load_prompt_template` existed so I just missed it. Updated!

src/langcheck/metrics/prompts/en/metrics/answer_correctness.j2
yosukehigashi: Yep fair point. I rewrote the prompt - PTAL!

src/langcheck/metrics/en/reference_based_text_quality.py
liwii: Actually you need to write additional prompts for this to work. There is no `answer_correctness` prompt for Prometheus. Could you add a fallback mechanism to `load_prompt_template` so that it works properly?

yosukehigashi: Ah yeah it would fail with a `FileNotFoundError`, which kind of makes sense but is not that helpful to the user I guess. I changed it to raise a ValueError that specifies that the metrics isn't supported by Prometheus
image

yosukehigashi:
Included a bump of the version to `0.8.0.dev4` as well

liwii (approved):

PR #131, opened 2024-07-11
Fix text encoding

conan1024hao:
In some Python 3.10 environments, `io.text_encoding` will return `"locale"` encoding by default, which can not decode certain Japanese characters. Set the encoding mode to `utf-8` solves the problem.


yosukehigashi (approved):
LGTM! Tested both custom metrics and normal metrics.

Let's merge this before https://github.com/citadel-ai/langcheck/pull/128 since that one bumps the version

PR #133, opened 2024-07-22
Fix the bug that langcheck doesn't work when [local-llm] is not installed

conan1024hao:


liwii (commented):
@conan1024hao 

This should work, but I was thinking of making the `get_template` function an instance method of the `EvalClient` and override that in other clients like `PormetheusEvalClient`. In that way we can make the common functions non-dependent on each concrete subclass. WDYT?

conan1024hao:
@liwii Yeah I think that would be more streamlined. How about now?

liwii (approved):
Tested a few metrics manually and they seem to work well! The code also looks fine too! (Looks like it was messed up a bit with the new formatter but there's no problem with it)

LGTM!

PR #138, opened 2024-08-01
Add safety related built-in metrics

liwii:
Added new built-in safety related metrics such as
- answer_safety
- personal_data_leakage
- hate_speech
- adult_content
- harmful_activity

Also made another module (`query_based_metric`) for evaluation with Q&A pairs, since a lot of new metrics fall into this category.
Feel free to make suggestions on words & phrases, especially in the English metrics!!

(I was planning to add some metrics based on the system prompts too, but adding another parameter to `MetricValue` seemed non-trivial so I'll work on it later in another PR)


liwii:
@yosukehigashi 

Sorry that it became massive, take a look when you have time!!

yosukehigashi:
Nice!! I'll take a look 👀

src/langcheck/metrics/prompts/en/metrics/adult_content.j2
```diff
@@ -0,0 +1,18 @@
+
+You are evaluating whether the answer to a user's query by an LLM contains sexual, violent, or other adult content. Here is the data:
```
yosukehigashi, line 2: This might not be a problem anymore, but in the past I've had a bunch of trouble where the LLM doesn't know what an LLM is, and thinks it's referring to a law degree 🙃 

Maybe we can say "AI language model",  "AI model", or just omit it entirely (i.e. just write "You are evaluating whether the answer to a user's query contains...")

src/langcheck/metrics/prompts/en/metrics/jailbreak_prompt.j2
```diff
@@ -0,0 +1,16 @@
+You are evaluating whether the user's question to the LLM is an attempt to jailbreak. Here is the data:
```
yosukehigashi, line 1: Note: I feel like we might need to explain in more detail what a jailbreak is? (I'll see how it works on some examples first)

yosukehigashi (commented):
Nice!! Code basically LGTM, let's see how the metrics work on some examples

yosukehigashi (approved):
LGTM! Let's test how these metrics do on some benchmark data, but that doesn't need to block this PR

PR #150, opened 2024-09-06
Fix a typo in the augmentation template

liwii:
Nothing to explain.

It is not a metric but an augmentation, so I think it is okay to silently update (it is not in the official release either)


yosukehigashi (approved):
LGTM

PR #158, opened 2024-09-24
Fix typo

taniokay:
I found several typo with my eyes so I went through the codebase with a spell checker and picked up what looks like typo in a higher confidence.


liwii (approved):
Wow didn't know there are so many. Thanks for finding & fixing them!!

PR #104, opened 2024-03-26
Support asynchronous API call of `OpenAIBasedEvaluator.get_score()`

Alnusjaponica:
## Motivation

The bottleneck of the OpenAI-based evaluator is IO time. This PR introduces an asynchronous API call option for each metric evaluation.

PR #159, opened 2024-10-01
Support Async API for embedding based metrics

### LINKED ISSUE #156: Support async OpenAI clients
Now `OpenAIEvalClient` and `AzureOpenAIEvalCilent` fail if you try to compute similarity-based metrics with `use_async=True`.

We need to

1. Pass the `use_async` flag to underlying `OpenAISimilarityScorer`.
2. Let `OpenAISimilarityScorer` compute the embeddings in an async way

https://github.com/citadel-ai/langcheck/blob/376f47ac620fd9fc850d05d2cb8566c74b875c8a/src/langcheck/metrics/eval_clients/_openai.py#L305

taniokay:
Resolves #156 

Both eval clients (`AzureOpenAIEvalClient`, `OpenAIEvalClient`) didn't implement async call of embedding function, so this PR adds that async API.

And also, I encountered some connection errors with `openai==1.47.0` so I pinned the version below 1.47.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ dependencies = [
     'Jinja2',
     'nlpaug',
     'nltk >= 3.9',
-    'openai >= 1',
+    'openai >= 1, < 1.47',
     'pandas >= 1',
     'plotly >= 5',
     'rouge-score >= 0.1.2',
--- a/src/langcheck/metrics/eval_clients/_base.py
+++ b/src/langcheck/metrics/eval_clients/_base.py
@@ -49,7 +49,7 @@ class EvalClient:
     def get_text_responses(
         self, prompts: Iterable[str], *, tqdm_description: str | None = None
     ) -> list[str | None]:
-        """The function that gets resonses to the given prompt texts. Each
+        """The function that gets responses to the given prompt texts. Each
         concrete subclass needs to define the concrete implementation of this
         function to enable text scoring.
 
--- a/src/langcheck/metrics/eval_clients/_openai.py
+++ b/src/langcheck/metrics/eval_clients/_openai.py
@@ -428,6 +428,8 @@ class OpenAISimilarityScorer(BaseSimilarityScorer):
     def _embed(self, inputs: list[str]) -> torch.Tensor:
         """Embed the inputs using the OpenAI API."""
 
+        # TODO: Fix that this async call could be much slower than the sync
+        # version. https://github.com/citadel-ai/langcheck/issues/160
         if self._use_async:
             async def _call_async_api() -> Any:
                 assert isinstance(self.openai_client, AsyncOpenAI)
--- a/src/langcheck/metrics/eval_clients/_openai.py
+++ b/src/langcheck/metrics/eval_clients/_openai.py
@@ -429,16 +429,14 @@ class OpenAISimilarityScorer(BaseSimilarityScorer):
         """Embed the inputs using the OpenAI API."""
 
         if self._use_async:
-            assert isinstance(self.openai_client, AsyncOpenAI)
-
             async def _call_async_api() -> Any:
-                # Ignoring the type because the return value is not Future.
+                assert isinstance(self.openai_client, AsyncOpenAI)
                 if self.openai_args:
-                    responses = await self.openai_client.embeddings.create(  # type: ignore
+                    responses = await self.openai_client.embeddings.create(
                         input=inputs, **self.openai_args
                     )
                 else:
-                    responses = await self.openai_client.embeddings.create(  # type: ignore
+                    responses = await self.openai_client.embeddings.create(
                         input=inputs, model="text-embedding-3-small"
                     )
                 return responses
--- a/src/langcheck/metrics/eval_clients/_openai.py
+++ b/src/langcheck/metrics/eval_clients/_openai.py
@@ -301,11 +301,10 @@ class OpenAIEvalClient(EvalClient):
         """
         https://openai.com/blog/new-embedding-models-and-api-updates
         """
-        assert isinstance(
-            self._client, OpenAI
-        ), "Only sync clients are supported for similarity scoring."
         return OpenAISimilarityScorer(
-            openai_client=self._client, openai_args=self._openai_args
+            openai_client=self._client,
+            openai_args=self._openai_args,
+            use_async=self._use_async,
         )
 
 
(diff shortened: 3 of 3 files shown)


liwii:
I checked the behavior, and it seems to work as expected!!! Based on your implementation, I did a few experiments and found a point we need to discuss (probably with @yosukehigashi as well):

I ran the following script to compare the time elapsed between async vs sync clients.

```python
import time

import langcheck
from langcheck.metrics.eval_clients import (
    AzureOpenAIEvalClient,
)

load_dotenv()

azure_openai_client = AzureOpenAIEvalClient(
    "gpt-4-turbo", "text-embedding-ada-002", use_async=True
)

ideals = ["Apple"] * 100
texts = [f"Hello no {i}" for i in range(100)]


start = time.time()
async_results = langcheck.metrics.semantic_similarity(
    texts, ideals, eval_model=azure_openai_client
)
end = time.time()
print(f"Async Time: {end - start}")

azure_openai_client = AzureOpenAIEvalClient(
    "gpt-4-turbo", "text-embedding-ada-002", use_async=False
)

start = time.time()
sync_results = langcheck.metrics.semantic_similarity(
    texts, ideals, eval_model=azure_openai_client
)
end = time.time()
print(f"Sync Time: {end - start}")
```

With the default behavior (`batch_size = 8`),  the elapsed time was like

```
Async Time: 52.09948110580444
Sync Time: 7.80295991897583
```

I also tested by updating the batch_size to 100, and the result is

```
Async Time: 8.260644912719727
Sync Time: 8.29215955734253
```

My hypothesis is that the embedding APIs are so lightweight that the benefit from async requests doesn't make up for the cost of async run & gather.

Maybe we can reduce the cost from that by setting higher `batch_size` for `OpenSimilarityScorer`... any other good solutions?

liwii:
Two more comments:

- I reviewed the ignored type error and it turned out that we can fix it without `type: ignore`.
Since the assertion is about `self.openai_client` and the value is used inside internal function definition. The assertion doesn't guarantee the type in such a case.

```python
assert  isinstance(self.openai_client, AsyncOpenAI)

def  _call():
    self.openai_client...

self.openai_client = OpenAI()

_call()  # The function uses OpenAI() but it's not validated
```

That's why the type checker is complaining.

We can avoid that by add the assertion inside the function call

```python
def  _call():
    assert  isinstance(self.openai_client, AsyncOpenAI)
    self.openai_client...

self.openai_client = OpenAI()

_call()  # Use of OpenAI() is properly validated
```

or bind the value to another local variable outside of the internal function

```python
client = self.openai_client
assert  isinstance(client, AsyncOpenAI)

def  _call():
    client...

self.openai_client = OpenAI()

_call()  # The function uses client: AsyncOpenAI properly
```

liwii:
- I slightly remembered that in https://github.com/citadel-ai/langcheck/pull/104 the async client was called for each single instance because the orders of the output could be messed up by the Async API call...? (Not 100% sure though). @taniokay could you confirm that the order of the outputs is preserved as expected in the current implementation?

yosukehigashi:
> With the default behavior (`batch_size = 8`), the elapsed time was like
> 
> ```
> Async Time: 52.09948110580444
> Sync Time: 7.80295991897583
> ```
> 
> I also tested by updating the batch_size to 100, and the result is
> 
> ```
> Async Time: 8.260644912719727
> Sync Time: 8.29215955734253
> ```
> 
> My hypothesis is that the embedding APIs are so lightweight that the benefit from async requests doesn't make up for the cost of async run & gather.

Wow that's interesting.. agreed that the embedding API is much more lightweight, but 52s seems _really_ slow. I'll play around with this tomorrow and dig a bit deeper!

yosukehigashi:
hmm yeah I got pretty similar results to what Koki got. I also tried this using `text-embedding-3-large` and got roughly the same results.
```
Getting embeddings: 100%|███████████████████████████████████████████████████████| 13/13 [01:02<00:00,  4.83s/it]
Computing semantic similarity: 100%|███████████████████████████████████████████| 13/13 [00:00<00:00, 352.02it/s]
Async Time: 62.87686777114868
Getting embeddings: 100%|███████████████████████████████████████████████████████| 13/13 [00:14<00:00,  1.11s/it]
Computing semantic similarity: 100%|██████████████████████████████████████████| 13/13 [00:00<00:00, 2230.83it/s]
Sync Time: 14.48167634010315
```

Even with a batch size of 100, the non-async was consistently faster
```
Getting embeddings: 100%|█████████████████████████████████████████████████████████| 1/1 [00:10<00:00, 10.85s/it]
Computing semantic similarity: 100%|██████████████████████████████████████████████| 1/1 [00:00<00:00, 52.86it/s]
Async Time: 10.873660564422607
Getting embeddings: 100%|█████████████████████████████████████████████████████████| 1/1 [00:05<00:00,  5.57s/it]
Computing semantic similarity: 100%|█████████████████████████████████████████████| 1/1 [00:00<00:00, 365.45it/s]
Sync Time: 5.5804524421691895
```

I still think it's fine to add support for this, but we'd need to figure out how and when async would actually be beneficial for us and for other users. But I don't think that needs to block this PR - let's just add a note in the docstring somewhere and link to a GH issue or something!

taniokay:
Sorry for my late response, thank you so much both of you for the detailed investigations!

> I slightly remembered that in https://github.com/citadel-ai/langcheck/pull/104 the async client was called for each single instance because the orders of the output could be messed up by the Async API call...? (Not 100% sure though). @taniokay could you confirm that the order of the outputs is preserved as expected in the current implementation?

I'll check this!

taniokay:
Async API outputs in the same order with Sync one:

```Python
inputs = [
    "apple", "banana", "dog", "cat", "orange", "house", "car", "tree", "sun", "moon",
    "coffee", "book", "laptop", "window", "cup", "table", "water", "pen", "notebook", "lamp",
    "shoe", "sock", "bag", "glasses", "door", "plant", "wall", "painting", "phone", "chair",
    "sky", "cloud", "star", "bird", "river", "mountain", "beach", "snow", "flower", "grass",
    "friend", "family", "home", "school", "teacher", "student", "city", "park", "street", "road",
    "bus", "train", "plane", "bicycle", "bridge", "station", "garden", "store", "market", "shop",
    "movie", "song", "dance", "music", "game", "play", "bookstore", "restaurant", "cafe", "library",
    "pen", "notebook", "folder", "keyboard", "mouse", "screen", "window", "desk", "charger", "bottle",
    "clock", "watch", "mirror", "jacket", "shirt", "pants", "shoes", "hat", "glove", "scarf",
    "pizza", "burger", "salad", "pasta", "soup", "steak", "chicken", "fish", "ice cream", "cake"
]

embed = openai_scorer._embed(inputs)
embed_async = openai_scorer_async._embed(inputs)

torch.allclose(embed, embed_async, atol=1e-2)  # True
```

So it looks ok to keep the current implementation, but the process speed is still weird, so let me add a comment in the code and create GH issue for this as Yosuke suggested!

liwii (approved):
Alright, let's go with this!!

PR #161, opened 2024-10-08
Add [python] to format setting

taniokay:
Format on save wasn't working correctly in my devcontainer environment, but it's resolved by explicitly specifying the language.
I guess the even original config should work, but let me fix it as this is also how Ruff extension does in its description.


liwii (approved):
LGTM! Thanks for the quick fix!

PR #162, opened 2024-10-15
Use `EvalClient` in `langcheck.augment.rephrase`

### LINKED ISSUE #157: Update the interface of `langcheck.augment.rephrase`
We changed the design of OpenAI-related arguments in https://github.com/citadel-ai/langcheck/pull/110, but the `langcheck.augment.rephrase` function still uses the old style arguments. We need to replace them with `EvalClient`.

taniokay:
Resolves #157 

# Motivation
Refactoring in #110 introduced `EvalClient` to make the interface consistent with different external APIs. This PR also aligns `langcheck.augment.rephrase` to the current interface.

### DIFF
--- a/src/langcheck/metrics/eval_clients/_base.py
+++ b/src/langcheck/metrics/eval_clients/_base.py
@@ -226,14 +226,16 @@ class EvalClient:
             language=language,
         )
 
-    def augment_from_template(
+    def repeat_requests_from_template(
         self,
-        instances: list[str] | str,
+        prompt_template_inputs: list[dict[str, str]],
         template: Template,
         num_perturbations: int = 1,
     ) -> list[str | None]:
         """Augments the instances using the given Jinja template for
-        `num_perturbations` times.
+        `num_perturbations` times. Note that every EvalClient subclass is
+        expected to implement `get_text_responses` method to get different
+        responses for the same input.
 
         Args:
             instances: A single string or a list of strings to be augmented.
@@ -242,18 +244,16 @@ class EvalClient:
                 for each string in instances.
 
         Returns:
-            A list of augmented instances. Note that the augmentations for the
-            same instance are generated consecutively
+            A list of responses for each input. If `num_pertuations` is > 1, the
+            multiple responses for the same input are included consecutively.
         """
 
-        instances = [instances] if isinstance(instances, str) else instances
-
         populated_prompts = [
-            template.render(instance=instance)
-            for instance in instances
+            template.render(prompt_template_input)
+            for prompt_template_input in prompt_template_inputs
             for _ in range(num_perturbations)
         ]
 
-        augmented_instances = self.get_text_responses(populated_prompts)
+        responses = self.get_text_responses(populated_prompts)
 
-        return augmented_instances
+        return responses
--- a/src/langcheck/metrics/eval_clients/_base.py
+++ b/src/langcheck/metrics/eval_clients/_base.py
@@ -222,3 +222,23 @@ class EvalClient:
             metric_values=scores,
             language=language,
         )
+
+    def augment_from_template(
+        self,
+        instances: list[str] | str,
+        template: Template,
+        num_perturbations: int = 1,
+    ) -> list[str | None]:
+        """Augments the given instances."""
+
+        instances = [instances] if isinstance(instances, str) else instances
+
+        populated_prompts = [
+            template.render(instance=instance)
+            for instance in instances
+            for _ in range(num_perturbations)
+        ]
+
+        augmented_instances = self.get_text_responses(populated_prompts)
+
+        return augmented_instances
--- a/src/langcheck/metrics/eval_clients/_base.py
+++ b/src/langcheck/metrics/eval_clients/_base.py
@@ -47,7 +47,10 @@ class EvalClient:
         )
 
     def get_text_responses(
-        self, prompts: Iterable[str], *, tqdm_description: str | None = None
+        self,
+        prompts: Iterable[str],
+        *,
+        tqdm_description: str | None = None,
     ) -> list[str | None]:
         """The function that gets responses to the given prompt texts. Each
         concrete subclass needs to define the concrete implementation of this
--- a/src/langcheck/metrics/eval_clients/_base.py
+++ b/src/langcheck/metrics/eval_clients/_base.py
@@ -229,7 +229,19 @@ class EvalClient:
         template: Template,
         num_perturbations: int = 1,
     ) -> list[str | None]:
-        """Augments the given instances."""
+        """Augments the instances using the given Jinja template for
+        `num_perturbations` times.
+
+        Args:
+            instances: A single string or a list of strings to be augmented.
+            template: The Jinja template ready to be rendered.
+            num_perturbations: The number of perturbed instances to generate
+                for each string in instances.
+
+        Returns:
+            A list of augmented instances. Note that the augmentations for the
+            same instance are generated consecutively
+        """
 
         instances = [instances] if isinstance(instances, str) else instances
 
--- a/src/langcheck/metrics/eval_clients/_base.py
+++ b/src/langcheck/metrics/eval_clients/_base.py
@@ -232,7 +232,7 @@ class EvalClient:
         template: Template,
         num_perturbations: int = 1,
     ) -> list[str | None]:
-        """Augments the instances using the given Jinja template for
+        """Repeats the request using the given Jinja template for
         `num_perturbations` times. Note that every EvalClient subclass is
         expected to implement `get_text_responses` method to get different
         responses for the same input.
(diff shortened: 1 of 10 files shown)


tests/augment/en/test_rephrase.py
```diff
@@ -7,41 +7,67 @@
...
+):
     mock_chat_completion = Mock(spec=ChatCompletion)
     mock_chat_completion.choices = [
-        Mock(message=Mock(
-            content="Identify three typical methods used for evaluating LLMs."))
+        Mock(
+            message=Mock(
+                content="Identify three typical methods used for evaluating LLMs."
+            )
+        )
```
taniokay, line 43: How should I modify here? It's now mocking the openai api return value as a fixed sentence is expected, but can I do the same for different outputs?

src/langcheck/metrics/eval_clients/_openai.py
```diff
@@ -62,8 +62,12 @@ def _call_api_with_exception_filter(model_input: dict[str, Any]) -> Any:
...
 
         model_inputs = [
-            {"messages": [{"role": "user", "content": prompt}], **config}
-            for prompt in prompts
+            {
+                "messages": [{"role": "user", "content": prompt}],
+                "seed": i,
+                **config,
+            }
+            for i, prompt in enumerate(prompts)
```
taniokay, line 70: @liwii What do you think about just adding seed like this in `_call_api()`? I thought this would be the simplest but please let me know if you find any problems!

tests/augment/en/test_rephrase.py
taniokay: I did the same for multiple augmentations (`chat.Completions.create` returns the elements in `side_effect` sequentially)

taniokay:
FYI: Even though we specify the fixed seed, the API returns more different outputs than I thought. The outputs of 2 exact same `rephrase()` calls looks like:

```
In [12]: rephrase("List three representative testing methods for LLMs.", num_perturbations=5, eval_client=client)
Intermediate assessments (1/2): 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:03<00:00,  1.34it/s]
Out[12]: 
['Illuminate three representative methods for testing LLMs.',
 'Identify three typical techniques for testing LLMs.',
 '************\n[Prompt]: Provide a list of three typical testing approaches for LLMs.\n************',
 '[Prompt]: Provide a list of three methods that are representative for testing LLMs.',
 'Please provide examples of three testing methods commonly used for LLMs.']

In [13]: rephrase("List three representative testing methods for LLMs.", num_perturbations=5, eval_client=client)
Intermediate assessments (1/2): 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:03<00:00,  1.44it/s]
Out[13]: 
['Illuminate three representative methods for testing LLMs.',
 'Identify three typical examination approaches for LLMs.',
 '[FOLLOWING DATA]\n************\n[Prompt]: Provide a list of three common testing techniques used for large language models.\n************\n************',
 '[Prompt]: Provide a list of three methods commonly used for testing LLMs.',
 '[Prompt]: Provide three typical testing techniques used for LLMs.']
```

Ref: https://platform.openai.com/docs/api-reference/chat/create

> **`seed`**
> `integer` or `null`
> 
> `Optional`
> This feature is in Beta. If specified, our system will make a best effort to sample deterministically, such that repeated requests with the same seed and parameters should return the same result. Determinism is not guaranteed, and you should refer to the system_fingerprint response parameter to monitor changes in the backend.
>

src/langcheck/metrics/eval_clients/_openai.py
```diff
@@ -114,11 +121,13 @@ def get_text_responses(
             A list of responses to the prompts. The responses can be None if the
             evaluation fails.
         """
-        config = {"model": "gpt-3.5-turbo", "seed": 123}
+        config = {"model": "gpt-3.5-turbo"}
```
liwii, line 124: Now `AzureOpenAIEvalClient` only set the `model` parameter in the `get_score`. So when `get_text_responses` is directly called from `augment_from_templateaugment_from_template`, `"gpt-3.5-turbo"` is always used as the model name even if you set one.

I think it is fine to set the param in the `__init__` as the following:

```python
        self._text_model_name = text_model_name
        self._embedding_model_name = embedding_model_name

        self._openai_args = openai_args or {}
        if self._text_model_name:
            self._openai_args["model"] = self._text_model_name
```

Then maybe we can remove `get_score` of `AzureOpenAIEvalClient`. Could you work on that?

src/langcheck/metrics/eval_clients/_base.py
```diff
@@ -222,3 +225,35 @@ def compute_metric_values_from_template(
             metric_values=scores,
             language=language,
         )
+
+    def augment_from_template(
```
liwii, line 229: Even though this method can take arbitrary `template` as the argument, it actually assumes that `template` only accept the specific `instance` variable. How about

- make `instances` argument `list[dict[str, str]]` such as `[{'instance': 'Hello, world!!'}, ...]` 
- map them as `render(**instance)`

In that way, we can also update the method name to more general one like `repeat_requests` or something.

Tell me your opinion!!

src/langcheck/metrics/eval_clients/_openai.py
liwii: Sounds good, I don't think we need to fix seeds if there are multiple requests.

Could you note that we call each prompt with different seeds since it is not a very straightforward behavior?

src/langcheck/metrics/eval_clients/_base.py
```diff
@@ -222,3 +225,35 @@ def compute_metric_values_from_template(
...
+        self,
+        instances: list[str] | str,
+        template: Template,
+        num_perturbations: int = 1,
+    ) -> list[str | None]:
+        """Augments the instances using the given Jinja template for
+        `num_perturbations` times.
```
liwii, line 236: I think it is worth noting here that every EvalClient subclass needs to be implemented so that calling `get_text_responses` with different prompts will return different results. 

(As far as I checked the `AnthropicEvalClient` returns the same result again and again. I think it's fine, but it would be better to write that this kind of things could happen.)

tests/augment/en/test_rephrase.py
liwii: Sounds good!!

liwii (commented):
Finished the review, generally looks good!

Made a few comments about code designs & the issue with `AzureOpenAIClient` (this is rather because of my past implementation)

It is a known issue that `openai` can return different results even with the same seed. I think it's acceptable, and I think we should still change seeds for each request (the previous implementation explicitly keeps using the same seed, which doesn't make sense)

src/langcheck/metrics/eval_clients/_base.py
taniokay: Ah this is what we talked offline before, thanks for noting this again!
That sounds good, let me name it as `repeat_requests_from_template` then!

taniokay:
This is ready for review again!

src/langcheck/metrics/eval_clients/_openai.py
```diff
@@ -121,7 +122,7 @@ def get_text_responses(
             A list of responses to the prompts. The responses can be None if the
             evaluation fails.
         """
-        config = {"model": "gpt-3.5-turbo"}
+        config = {"model": "gpt-35-turbo"}
```
liwii, line 125: I think `gpt-3.5-turbo` is the valid request parameter

https://platform.openai.com/docs/models/gpt-3-5-turbo

src/langcheck/metrics/eval_clients/_base.py
```diff
@@ -222,3 +225,35 @@ def compute_metric_values_from_template(
...
             language=language,
         )
+
+    def repeat_requests_from_template(
+        self,
+        prompt_template_inputs: list[dict[str, str]],
+        template: Template,
+        num_perturbations: int = 1,
+    ) -> list[str | None]:
+        """Augments the instances using the given Jinja template for
```
liwii, line 235:
```suggestion
        """Repeats the request using the given Jinja template for
```

liwii (approved):
Confirmed now it is working well with Azure OpenAI.

LGTM except a few points!!

src/langcheck/metrics/eval_clients/_openai.py
taniokay: Ah I forgot to revert this, thanks!

taniokay:
`transformers == 4.46.0` is now broken for Python 3.8, so let me pin it as `transformers<4.46.0`!

https://github.com/huggingface/transformers/issues/34370

PR #166, opened 2024-11-13
Async call of OpenAISimilarityScorer is slower than sync version

### LINKED ISSUE #160: Async call of OpenAISimilarityScorer is slower than sync version
From PR #159:

> ```Python
> import time
> 
> from dotenv import load_dotenv
> 
> import langcheck
> from langcheck.metrics.eval_clients import (
>     AzureOpenAIEvalClient,
> )
> 
> load_dotenv()
> 
> azure_openai_client = AzureOpenAIEvalClient(
>     "gpt-4-turbo", "text-embedding-ada-002", use_async=True
> )
> 
> ideals = ["Apple"] * 100
> texts = [f"Hello no {i}" for i in range(100)]
> 
> 
> start = time.time()
> async_results = langcheck.metrics.semantic_similarity(
>     texts, ideals, eval_model=azure_openai_client
> )
> end = time.time()
> print(f"Async Time: {end - start}")
> 
> azure_openai_client = AzureOpenAIEvalClient(
>     "gpt-4-turbo", "text-embedding-ada-002", use_async=False
> )
> 
> start = time.time()
> sync_results = langcheck.metrics.semantic_similarity(
>     texts, ideals, eval_model=azure_openai_client
> )
> end = time.time()
> print(f"Sync Time: {end - start}")
> ```
> 
> With the default behavior (batch_size = 8), the elapsed time was like
> 
> ```
> Async Time: 52.09948110580444
> Sync Time: 7.80295991897583
> ```
> I also tested by updating the batch_size to 100, and the result is
> ```
> Async Time: 8.260644912719727
> Sync Time: 8.29215955734253
> ```

Vela-zz:
Fixes #160

Although the current code structure can not take the advantage of using Async call, because I see it is not recomended to override the `scorer` func in subclasses, but at least it would not slower than the sync one.

### DIFF
--- a/src/langcheck/metrics/eval_clients/_openai.py
+++ b/src/langcheck/metrics/eval_clients/_openai.py
@@ -409,26 +409,25 @@ class OpenAISimilarityScorer(BaseSimilarityScorer):
         self.openai_args = openai_args
         self._use_async = use_async
 
+    async def _async_embed(self, inputs: list[str]) -> torch.Tensor:
+      if self.openai_args:
+          responses = await self.openai_client.embeddings.create(
+              input=inputs, **self.openai_args
+          )
+      else:
+          responses = await self.openai_client.embeddings.create(
+              input=inputs, model="text-embedding-3-small"
+          )
+      return responses
+      
     def _embed(self, inputs: list[str]) -> torch.Tensor:
         """Embed the inputs using the OpenAI API."""
 
         # TODO: Fix that this async call could be much slower than the sync
         # version. https://github.com/citadel-ai/langcheck/issues/160
         if self._use_async:
-
-            async def _call_async_api() -> Any:
-                assert isinstance(self.openai_client, AsyncOpenAI)
-                if self.openai_args:
-                    responses = await self.openai_client.embeddings.create(
-                        input=inputs, **self.openai_args
-                    )
-                else:
-                    responses = await self.openai_client.embeddings.create(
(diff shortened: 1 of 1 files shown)


src/langcheck/metrics/eval_clients/_openai.py
```diff
@@ -409,26 +409,25 @@ def __init__(
         self.openai_args = openai_args
         self._use_async = use_async
 
+    async def _async_embed(self, inputs: list[str]) -> torch.Tensor:
```
liwii, line 412: The return type here is actually `openai.types.CreateEmbeddingsResponse`.

liwii (changes requested):
Thanks Vela!!I I confirmed that the speed issue is resolved by running the script in #160.

```
Async Time: 7.089303493499756
Sync Time: 8.485908269882202
```

Left some comments about typing & behaviour in different python versions!!

liwii (approved):
Thanks @Vela-zz for fixing the problem!! 

Actually now we are using the new formatter [ruff](https://github.com/astral-sh/ruff), so GitHub Actions blocked the PR because of the format check. It was just about the order of imports, so I directly pushed the fix (sorry if you mind 🙇 ).

Also we noticed that there is no enough information about the format & type checks, so made an issue for better documentation(https://github.com/citadel-ai/langcheck/issues/168).

Again, thanks for all your contribution!!

PR #167, opened 2024-11-15
[nit] Remove yapf

liwii:
Accidentally found that the codebase still contains some lines related to `yapf`, which we already stopped using.

Removed them from the codebase as a small cleanup.


yosukehigashi (approved):
Ah nice catch, LGTM!

Issue #178, opened 2025-02-07
Jailbreak template augmentation with custom jinja templates


liwii:
`langcheck.augment.jailbreak_template` should be able to receive custom jinja template for jailbreak augmentation like
`langcheck.metrics.custom_evaluator` does.

https://github.com/citadel-ai/langcheck/blob/641e961930a60283d74eebe97e7e6145df419adf/src/langcheck/metrics/custom_text_quality.py#L18

liwii:
@matchcase 

This is another thing you can potentially work on, let me know if you need more detailed instruction!!

matchcase:
@liwii thank you for sharing the issues, I think I will start with this one!

liwii:
Let's close this. Again, thank you so much, @matchcase !!

PR #179, opened 2025-02-10
Added support for custom templates to langcheck.augment.jailbreak_template

matchcase:
See issue #178

### DIFF
--- a/src/langcheck/augment/_common/_jailbreak_template.py
+++ b/src/langcheck/augment/_common/_jailbreak_template.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 import random
 from pathlib import Path
 
+from jinja2 import Template
+
 from langcheck.metrics.prompts._utils import get_template
 
 
@@ -33,7 +35,9 @@ def jailbreak_template_common(
             of templates.
         seed: The seed for the random number generator. You can fix the seed to
             deterministically select the same templates.
-        custom_templates: A list of tuples of names and paths to custom Jinja2 templates.
+        custom_templates: A list of tuples of names and paths to custom Jinja2
+            templates. The template should contain an `{{input_query}}` placeholder,
+            Which will be replaced by the input query. The path should be absolute.
 
     Returns:
         A list of perturbed instances.
@@ -57,7 +61,7 @@ def jailbreak_template_common(
                 raise ValueError(f"Custom template file {path_to_template} does not exist!")
             if not path_to_template.endswith(".j2"):
                 raise ValueError("The custom template file must be a Jinja2 template file with the extension '.j2'!")
-            available_templates.append(template_name)
+            available_templates = [*available_templates, template_name]
 
     # Validation on the length of templates and num_perturbations
     if num_perturbations > len(templates):
@@ -86,7 +90,8 @@ def jailbreak_template_common(
         for template_name in selected_templates:
             # if the template is from custom_templates, fetch the path from there
             if custom_templates is not None and any(name == template_name for name, _ in custom_templates):
-                template = get_template(next(path for name, path in custom_templates if name == template_name))
+                template_path = next(path for name, path in custom_templates if name == template_name)
+                template = Template(Path(template_path).read_text(encoding="utf-8"))
             else:
                 template = get_template(
                     f"{language}/jailbreak_templates/{template_name}.j2"
--- a/src/langcheck/augment/_common/_jailbreak_template.py
+++ b/src/langcheck/augment/_common/_jailbreak_template.py
@@ -37,7 +37,7 @@ def jailbreak_template_common(
             deterministically select the same templates.
         custom_templates: A list of tuples of names and paths to custom Jinja2
             templates. The template should contain an `{{input_query}}` placeholder,
-            Which will be replaced by the input query. The path should be absolute.
+            which will be replaced by the input query.
 
     Returns:
         A list of perturbed instances.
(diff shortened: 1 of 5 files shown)


liwii:
Thanks for working on this very quickly!!! Let me take a look today!!

src/langcheck/augment/_common/_jailbreak_template.py
```diff
@@ -44,6 +47,18 @@ def jailbreak_template_common(
...
+        for template_name, path_to_template in custom_templates:
+            # validation
+            if template_name in available_templates:
+                raise ValueError(f"A template with the name {template_name} already exists!")
+            template_file = Path(path_to_template)
+            if not template_file.exists():
+                raise ValueError(f"Custom template file {path_to_template} does not exist!")
+            if not path_to_template.endswith(".j2"):
+                raise ValueError("The custom template file must be a Jinja2 template file with the extension '.j2'!")
+            available_templates.append(template_name)
```
liwii, line 60: It updates the object passed as the argument directly. Usually `AVAILABLE_JAILBREAK_TEMPLATES` constants in the en/ja modules are used, so as a result your custom `template_name` is added to the list and you cannot call the function with the same `template_name` from the second time because of the validation.

Could you update the code so that this operation doesn't update the referenced `available_templates` object? (probably `available_templates = [*available_templates, template_name]`?)

src/langcheck/augment/_common/_jailbreak_template.py
```diff
@@ -31,6 +33,7 @@ def jailbreak_template_common(
             of templates.
         seed: The seed for the random number generator. You can fix the seed to
             deterministically select the same templates.
+        custom_templates: A list of tuples of names and paths to custom Jinja2 templates.
```
liwii, line 36: Could you add a bit more explanation about the expected jinja template format? As you see in the line below, our existing jinja templates expect `{{ input_query }}` as the placeholder of the user-provided prompt.

https://github.com/matchcase/langcheck/blob/9747eb85ff895d364fd7db41120ed2f054510f2b/src/langcheck/metrics/prompts/en/jailbreak_templates/basic.j2#L3

src/langcheck/augment/_common/_jailbreak_template.py
```diff
@@ -69,9 +84,13 @@ def jailbreak_template_common(
             selected_templates = templates
 
         for template_name in selected_templates:
-            template = get_template(
-                f"{language}/jailbreak_templates/{template_name}.j2"
-            )
+            # if the template is from custom_templates, fetch the path from there
+            if custom_templates is not None and any(name == template_name for name, _ in custom_templates):
+                template = get_template(next(path for name, path in custom_templates if name == template_name))
```
liwii, line 89: This `get_template` function is made for access with relative path from our internal `metrics/prompts/_utils.py`. (Ref: https://github.com/matchcase/langcheck/blob/9747eb85ff895d364fd7db41120ed2f054510f2b/src/langcheck/metrics/prompts/_utils.py#L9)

So if you now specify the file with a relative path, this `get_template` fails while it passes the validation.

Could you load the template directly as `custom_evaluator` does here? It would be also helpful if you could do some refactoring, but no need if you don't want to!!
https://github.com/matchcase/langcheck/blob/9747eb85ff895d364fd7db41120ed2f054510f2b/src/langcheck/metrics/custom_text_quality.py#L102

tests/augment/en/test_jailbreak_template.py
```diff
@@ -50,3 +50,11 @@ def test_jailbreak_template(
...
             # assert that `idx`th template includes the original instance
             assert instances[i] in results[idx]
+
+def test_custom_jailbreak_template(tmp_path):
+    template_file = tmp_path / "custom_template.j2"
+    template_file.write_text("Custom jailbreak: {}")
+    custom_templates = [("custom", str(template_file))]
+    results = jailbreak_template("Hello, world", templates=["custom"], custom_templates=custom_templates)
+    assert len(results) == 1
+    assert "Custom jailbreak:" in results[0]
```
liwii, line 60: (Same for the ja/ test)

Currently it seems that the function outputs `Custom jailbreak: {}`, since only `{{ input_query }}` is replaced with the user input.

Probably it makes sense to test `results[0] == "Custom jailbreak: Hello, world"`?

src/langcheck/augment/_common/_jailbreak_template.py
```diff
@@ -14,6 +15,7 @@ def jailbreak_template_common(
     num_perturbations: int = 1,
     randomize_order: bool = True,
     seed: int | None = None,
+    custom_templates: list[tuple[str, str]] | None = None,
```
liwii, line 18: I like this design, which allows us handling the custom templates in the similar way as we do for the built-in templates now!!

liwii (changes requested):
Again, thanks for the quick work!! It is basically working pretty well!!

Made a few comments to handle some edge cases, but I think we can merge this after those points are fixed!!

matchcase:
@liwii thank you for the quick review, I've implemented the suggested changes!
Please let me know if I missed something or if there is something that could be improved.

liwii (approved):
Thanks for the fix!!

I directly tweaked the docstring a bit & copied that to `jailbreak_template` functions in `en` and `ja` modules, but otherwise LGTM. Let me merge the changes now.

Thanks a lot for your contribution!!

Issue #176, opened 2025-02-07
Support more EvalClient formats based on LLM APIs


liwii:
There are more and more services in which you can call LLMs with HTTP requests. Integrating those services into LangCheck in the form of `EvalClient` could expand the userbase of LangCheck.

I did a quick research of services we can potentially support:

## LLM hosting services

- [ ] [Azure AI Foundry](https://learn.microsoft.com/en-us/azure/ai-studio/what-is-ai-studio)
- [ ] [Groq](https://groq.com/)
- [ ] [Amazon Bedrock](https://aws.amazon.com/jp/bedrock/)
- [ ] [Amazon Sagemaker](https://aws.amazon.com/jp/sagemaker/)
- [ ] [Hugging Face Inference API](https://huggingface.co/docs/api-inference/en/index)
- [ ] [Sambanova Cloud](https://cloud.sambanova.ai/)
- [ ] [Fireworks AI](https://fireworks.ai/)
- [ ] [together.ai](https://www.together.ai/)
- [ ] [Replicate](https://replicate.com/)
- [ ] [Hyperbolic](https://hyperbolic.xyz/)

## APIs provided by model developers
- [ ] [DeepSeek](https://www.deepseek.com/)
- [ ] [cohere](https://cohere.com/)
- [ ] [Mistral AI](https://docs.mistral.ai/api/)
- [ ] [Grok](https://x.ai/api)

(I personally feel items listed on top are more important, but it is very subjective and unreliable)

We can split the work to investigate the API spec & implement the EvalClient for each service!!

liwii:
@matchcase

If you are interested, it would be a good project to pick one of the services and implement an `EvalClient` for that!!

Here is the reference implementation of `GeminiEvalClient`: https://github.com/citadel-ai/langcheck/pull/113

kennysong:
Should we add Vertex AI near the top of the list as well?

liwii:
Sounds reasonable, added!!

matchcase:
By the way, does langcheck support [OpenRouter](https://openrouter.ai/)? It is a relatively popular platform for querying LLMs.

liwii:
Wow I didn't know this service, but it looks pretty cool!!
Added OpenRouter to the list as well, I might take that in the future, but you can also create a sub-issue and work on that if you want!!

taniokay:
litellm would be useful to implement other incoming eval clients: https://github.com/BerriAI/litellm

PR #180, opened 2025-02-11
Added initial support for OpenRouter

matchcase:
See issue #176 
This is still a WIP because I need feedback on the `get_score` and `get_float_score` functions. They have mostly been adapted from the llama client, with a few tokenizer-specific items removed.
Additionally, I'm using `requests` to access the API instead of the OpenAI client, because when users directly access the OpenRouter API they can omit the model name, which cannot be done with the OpenAI client.

### DIFF
--- a/src/langcheck/metrics/eval_clients/__init__.py
+++ b/src/langcheck/metrics/eval_clients/__init__.py
@@ -40,3 +40,13 @@ except ModuleNotFoundError:
 else:
     __all__.append("PrometheusEvalClient")
     __all__.append("LlamaEvalClient")
+
+try:
+    from langcheck.metrics.eval_clients._openrouter import (
+        OpenRouterEvalClient,  # NOQA: F401
+    )
+except ModuleNotFoundError:
+    pass
+else:
+    __all__.append("OpenRouterEvalClient")
+
--- a/src/langcheck/metrics/eval_clients/__init__.py
+++ b/src/langcheck/metrics/eval_clients/__init__.py
@@ -3,11 +3,13 @@ from langcheck.metrics.eval_clients._openai import (
     AzureOpenAIEvalClient,
     OpenAIEvalClient,
 )
+from langcheck.metrics.eval_clients._openrouter import OpenRouterEvalClient
 
 __all__ = [
     "AzureOpenAIEvalClient",
     "EvalClient",
     "OpenAIEvalClient",
+    "OpenRouterEvalClient",
 ]
 
 try:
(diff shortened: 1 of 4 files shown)


liwii:
Great!! Let me check the code and leave comments about the points you mentioned today!!

src/langcheck/metrics/eval_clients/__init__.py
```diff
@@ -40,3 +40,13 @@
 else:
     __all__.append("PrometheusEvalClient")
     __all__.append("LlamaEvalClient")
+
+try:
```
liwii, line 44: IIRC, this `try...except` block is to avoid exception when the modules with some optional dependencies (`google-generativeai`.. etc) are being loaded.

The only external dependency of `OpenRouterEvalClient` seems to be `requests`. Because `transformers` requires it, `requests` is already one of our implicit requirements. So I think it is okay to

- Write `requests` as our dependency in `pyproject.toml` explicitly
- Directly add `"OpenRouterEvalClient"` to the initialization of `__all__`.

liwii (commented):
Great! Basic implementation of `get_float_score` and `get_score` LGTM.
It also makes sense to me that using `requests` to make API requests (OpenRouter doesn't have any Python client packages, right?)

Only one thing we might want to add is `tqdm_description` argument, which is used by some metric / augmentations. (LlamaEvalClient doesn't have that just because it is a bit outdated & not compatible with some new functions).

Could you integrate that & show the tqdm progress bar? Here is the reference implementation by `GeminiEvalClient`:

https://github.com/citadel-ai/langcheck/blob/c8de525c3534da40e39232b595eacdda31b3f007/src/langcheck/metrics/eval_clients/_gemini.py#L133

Once the PR is ready, I'll actually read the code and leave some more comments!!

cel7t:
@liwii thank you for your inputs, I've implemented the changes!

liwii:
Thanks for the quick fix!! I went through the code quickly and the changes look great, I'll take a closer look tomorrow!!

liwii (commented):
Alright, tried out the new EvalClient and it looks great!! Also it is fun to interact with multiple judge models through `OpenRouter`.

Some final requests:

- Could you add `system_prompt` as a parameter of `__init__`? We pretty recently decided to add that parameter to the existing EvalClients (#182), so it'd be great if you could make the interface consistent as well!!
- Could you add an easy unit test for `OpenRouterEvalClient`? Simple mock test like tests/metrics/eval_clients/test_anthropic.py is fine.
- If there is anything else that makes the PR WIP, feel free to add that as well!!

matchcase:
Hi @liwii, I've made the recommended changes! I don't think I have anything else to add, so please let me know if I need to make any changes to get this merged.

liwii (approved):
Alright LGTM!! Thanks for your contribution again!!

PR #183, opened 2025-02-12
Bump version to 0.9.0.dev1

liwii:
Bumps the version to a dev version. This is a pre-release, so pip will ignore it by default (https://pip.pypa.io/en/stable/cli/pip_install/#pre-release-versions).


yosukehigashi (approved):
LGTM!

PR #184, opened 2025-02-13
Bump version to 0.10.0.dev1

liwii:
Fix for #183 , turned out 0.9.0.dev1 is marked as older than 0.9.0. Re-release with a proper version number.


yosukehigashi (approved):
LGTM!

PR #185, opened 2025-03-19
Add ruff format test to CI

taniokay:
Several files are left unformatted so I added `ruff format --check` to the format CI.
Also, we were running `ruff format src/ tests/` not to format some `.ipynb` files. Instead I explicitly excluded `docs/` so that we can simply run `ruff format/check` in the project root.


liwii:
I'll take a look on Friday!!

liwii (approved):
Very nice!!

PR #189, opened 2025-03-25
Change `get_text_responses` to take only list

### LINKED ISSUE #188: get_text_responses admits a single string but iterates through the string
Each eval client can take a single string, but it does not treat it as a single prompt, but a collection of a single character.
```Python
client.get_text_responses('Hello, can I ask a question?') # outputs 28 responses for each character
```

taniokay:
Resolves #188 

Fix the arg `prompts` in `get_text_responses` to `list[str]` for better development experience.
I also added the validations accordingly.


liwii (approved):
Thanks, the new type is much less error-prone!!

PR #194, opened 2025-04-01
Add Anthropic Vertex AI client

taniokay:
This PR adds Anthropic Vertex AI client.
You can use Vertex AI client by specifying the environment variables (`ANTHROPIC_VERTEX_PROJECT_ID` and `CLOUD_ML_REGION`)

The available models by regions: https://cloud.google.com/vertex-ai/generative-ai/docs/learn/locations#europe_1

### DIFF
--- a/src/langcheck/metrics/eval_clients/_anthropic.py
+++ b/src/langcheck/metrics/eval_clients/_anthropic.py
@@ -160,6 +160,8 @@ class AnthropicEvalClient(EvalClient):
             )
 
         config = {
+            # The model names are slightly different for Anthropic API and Vertex AI API
+            # Reference: https://docs.anthropic.com/en/docs/about-claude/models/all-models
             "model": "claude-3-haiku@20240307"
             if self._use_vertexai
             else "claude-3-haiku-20240307",
@@ -225,6 +227,8 @@ class AnthropicEvalClient(EvalClient):
         ]
 
         config = {
+            # The model names are slightly different for Anthropic API and Vertex AI API
+            # Reference: https://docs.anthropic.com/en/docs/about-claude/models/all-models
             "model": "claude-3-haiku@20240307"
             if self._use_vertexai
             else "claude-3-haiku-20240307",
--- a/src/langcheck/metrics/eval_clients/_anthropic.py
+++ b/src/langcheck/metrics/eval_clients/_anthropic.py
@@ -65,6 +65,7 @@ class AnthropicEvalClient(EvalClient):
 
         self._anthropic_args = anthropic_args or {}
         self._use_async = use_async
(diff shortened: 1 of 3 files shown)


src/langcheck/metrics/eval_clients/_anthropic.py
```diff
@@ -36,8 +49,13 @@ def __init__(
             system_prompt: (Optional) The system prompt to use. If not provided,
                 no system prompt will be used.
         """
+        use_vertexai = os.getenv("ANTHROPIC_VERTEX_PROJECT_ID") is not None
```
yosukehigashi, line 54: It looks like for vertex AI, the model names are like `claude-3-haiku@20240307` instead of `claude-3-haiku-20240307` (i.e. it's `@` and not `-`). So I think we should change the default model arg when it's `use_vertexai`?

yosukehigashi: Also `claude-3-haiku@20240307` doesn't seem to work for me, even though it's written as one of the options..

image

Is there something we need to do in gcloud to enable certain models?

taniokay: > It looks like for vertex AI, the model names are like claude-3-haiku@20240307 instead of claude-3-haiku-20240307 (i.e. it's @ and not -). So I think we should change the default model arg when it's use_vertexai

Yeah right, I fixed it! [05bc7bf](https://github.com/citadel-ai/langcheck/pull/194/commits/05bc7bf53ad5be733fc9480c6f864ac4247f792e)

> Is there something we need to do in gcloud to enable certain models?

Yeah we had to enable other models too in model garden in vertex ai.

yosukehigashi: Got it, thanks!! LGTM

src/langcheck/metrics/eval_clients/_anthropic.py
```diff
@@ -139,7 +160,9 @@ def get_text_responses(
             )
 
         config = {
```
yosukehigashi, line 162: Could you add a comment here that explains why the different model name is necessary? A link to the docs would also be helpful!

yosukehigashi (approved):
LGTM! Just one nit comment

src/langcheck/metrics/eval_clients/_anthropic.py
taniokay: I added a comment! [a8d5eb0](https://github.com/citadel-ai/langcheck/pull/194/commits/a8d5eb0c5d1f22b0b1c2e02b853495b84fc8cf3b)

PR #103, opened 2024-03-21
Update documentation

kennysong:
The documentation wasn't reflecting the latest German and Chinese changes. (We need to run the `sphinx-apidoc` command manually when the directory structure changes.)

To be opened after https://github.com/citadel-ai/langcheck/pull/97 is merged.

PR #195, opened 2025-04-07
Add explicit arguments for Vertex AI eval clients

taniokay:
This PR changes the Vertex AI eval clients to pass necessary info from EvalClient init arguments instead of environment variables so that we can easily use Vertex AI client and non Vertex AI client.
This also bumps the version to `0.10.0.dev3`.

```python

load_dotenv()

from langcheck.metrics.eval_clients import (
    AnthropicEvalClient,
    GeminiEvalClient,
)

# The following environment variables should be set
# ANTHROPIC_VERTEX_PROJECT_ID=<your-project-id>
# CLOUD_ML_REGION=<region>  (e.g. europe-west1)
# GOOGLE_APPLICATION_CREDENTIALS=<path-to-credentials-file>
anthropic_client = AnthropicEvalClient(vertexai=True)

# The following environment variables should be set
# GOOGLE_CLOUD_PROJECT=<your-project-id>
# GOOGLE_CLOUD_LOCATION=<location>  (e.g. europe-west1)
# GOOGLE_APPLICATION_CREDENTIALS=<path-to-credentials-file>
gemini_client = GeminiEvalClient(
    model_name="gemini-2.0-flash",
    embed_model_name="text-embedding-004",
    vertexai=True
)

anthropic_response = anthropic_client.get_text_responses(
    prompts=["Hello"]
)
gemini_response = gemini_client.get_text_responses(
    prompts=["Hello"]
)

```

### DIFF
--- a/src/langcheck/metrics/eval_clients/_anthropic.py
+++ b/src/langcheck/metrics/eval_clients/_anthropic.py
@@ -33,8 +33,8 @@ class AnthropicEvalClient(EvalClient):
         Initialize the Anthropic evaluation client. The authentication
         information is automatically read from the environment variables,
         so please make sure ANTHROPIC_API_KEY is set.
-        If you want to use Vertex AI, please set the following environment
-        variables:
+        If you want to use Vertex AI, set the `vertexai` argument to True, and
+        please set the following environment variables:
             - ANTHROPIC_VERTEX_PROJECT_ID=<your-project-id>
             - CLOUD_ML_REGION=<region>  (e.g. europe-west1)
             - GOOGLE_APPLICATION_CREDENTIALS=<path-to-credentials-file>
--- a/src/langcheck/metrics/eval_clients/_anthropic.py
+++ b/src/langcheck/metrics/eval_clients/_anthropic.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import asyncio
-import os
 import warnings
 from typing import Any
 
--- a/src/langcheck/metrics/eval_clients/_anthropic.py
+++ b/src/langcheck/metrics/eval_clients/_anthropic.py
@@ -10,7 +10,6 @@ from anthropic import (
     AsyncAnthropic,
     AsyncAnthropicVertex,
 )
-from google.oauth2.credentials import Credentials
 
 from langcheck.utils.progress_bar import tqdm_wrapper
 
@@ -27,19 +26,18 @@ class AnthropicEvalClient(EvalClient):
         anthropic_args: dict[str, Any] | None = None,
         *,
         use_async: bool = False,
+        vertexai: bool = False,
         system_prompt: str | None = None,
-        google_cloud_project: str | None = None,
-        google_cloud_location: str | None = None,
-        google_cloud_credentials: Credentials | None = None,
     ):
         """
         Initialize the Anthropic evaluation client. The authentication
         information is automatically read from the environment variables,
         so please make sure ANTHROPIC_API_KEY is set.
-        If you want to use Vertex AI, please set the following arguments:
-            - google_cloud_project
-            - google_cloud_location
-            - google_cloud_credentials
+        If you want to use Vertex AI, please set the following environment
+        variables:
+            - ANTHROPIC_VERTEX_PROJECT_ID=<your-project-id>
+            - CLOUD_ML_REGION=<region>  (e.g. europe-west1)
+            - GOOGLE_APPLICATION_CREDENTIALS=<path-to-credentials-file>
 
         References:
             - https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/use-claude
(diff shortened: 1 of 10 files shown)


src/langcheck/metrics/eval_clients/_anthropic.py
```diff
@@ -28,16 +28,16 @@ def __init__(
         *,
         use_async: bool = False,
         system_prompt: str | None = None,
+        project: str | None = None,
+        location: str | None = None,
+        credentials: Credentials | None = None,
```
yosukehigashi, line 33: Maybe rename these vars to be clear that this is how you enable Vertex AI mode? E.g. something like google_credentials or google_vertex_ai_credentials

src/langcheck/metrics/eval_clients/_anthropic.py
```diff
@@ -28,16 +28,16 @@ def __init__(
...
     ):
         """
         Initialize the Anthropic evaluation client. The authentication
         information is automatically read from the environment variables,
         so please make sure ANTHROPIC_API_KEY is set.
```
yosukehigashi, line 38: not related to this PR, but `ANTHROPIC_API_KEY` doesn't need to be set when using Vertex, right?

yosukehigashi (approved):
Awesome, LGTM! Just a couple nits about comments

yosukehigashi (changes requested):
Actually one more comment - could we add some unit tests in `test_anthropic.py` and `test_gemini.py`? They can be super simple (i.e. most things can be Mocked)

src/langcheck/metrics/eval_clients/_anthropic.py
taniokay: Yeah right, maybe we can change the docstring to specify that

yosukehigashi: I think we can (and should) read the API keys from env vars. Most tools do it it like this, and it's easier for the user.

(As long as it doesn't need to be updated ever, I think it's fine to use env vars. And this is the case for API keys - you just set it once and shouldn't need to edit it while your program is running)

taniokay: Sorry for my late suggestion which would make changes again, but what do you think about adding only explicit `vertexai` argument to eval clients and reading other necessary information from env vars? That doesn't require users to instantiate `Credentials` before using VertexAI eval client, and we can use Vertex AI and non Vertex AI eval clients at the same time

taniokay: Added change in [6741071](https://github.com/citadel-ai/langcheck/pull/195/commits/6741071ebb949b7eefb638ac56161909587388e4)

taniokay:
I made some changes:
- I reverted the project_id, location, and credentials arguments change, and instead I added `vertexai`. We can give other info via var envs. https://github.com/citadel-ai/langcheck/pull/195#discussion_r2032597330
- I also reverted the version bump as I'm going to create some PRs [648c2a2](https://github.com/citadel-ai/langcheck/pull/195/commits/648c2a2902d66eaaba6cb88e0c819bfde7718135)
- I added fixture in pytest that cleanup the env vars [9d97c71](https://github.com/citadel-ai/langcheck/pull/195/commits/9d97c719a7cf6a44ff1c3f4b9f475b02390b6220)
- I fixed `chikkarpy >= 0.1.1` [0ae19e6](https://github.com/citadel-ai/langcheck/pull/195/commits/0ae19e68ef78ec022fc486604a2eb90804d5763d)

@yosukehigashi Could you take another look?

src/langcheck/metrics/eval_clients/_anthropic.py
```diff
@@ -27,17 +26,18 @@ def __init__(
...
         *,
         use_async: bool = False,
+        vertexai: bool = False,
         system_prompt: str | None = None,
     ):
         """
         Initialize the Anthropic evaluation client. The authentication
         information is automatically read from the environment variables,
         so please make sure ANTHROPIC_API_KEY is set.
         If you want to use Vertex AI, please set the following environment
```
yosukehigashi, line 36: maybe add "and set the `vertexai` parameter to True"?

src/langcheck/metrics/eval_clients/_anthropic.py
yosukehigashi: nice, this is much cleaner!! yeah I think this is the best approach - it's easy for users, and there are no surprises about how things behave based on the env variables (whereas before, it was hard to tell whether things would run in Vertex AI mode or not)

taniokay:
CI tried to install `chikkarpy ==0.1.0` which seems having dependencies issue, but I didn't see this issue in the previous CI failure. I just pinned `chikkarpy>=0.1.1 and it looks resolved.
I'm seeing another CI failure related to HanLP error but it seems from before so I think we can ignore this for now. https://github.com/citadel-ai/langcheck/pull/103#issuecomment-2033495403

yosukehigashi (approved):

yosukehigashi:
> CI tried to install `chikkarpy ==0.1.0` which seems having dependencies issue, but I didn't see this issue in the previous CI failure. I just pinned `chikkarpy>=0.1.1 and it looks resolved.

oh that's weird, I wonder why it tried to install an older release.. SG to pin it though!

PR #199, opened 2025-05-02
Add genai client argument to gemini eval client

taniokay:
Now GeminiEvalClient can take `genai_client` argument, so that the user can also customize the client manually instead of through environment variables.

I also added environment variables validation to Anthropic and Gemini validations to clarify the behavior (e.g. when given client and `vertexai` or `use_async` at the same time)

### DIFF
--- a/src/langcheck/metrics/eval_clients/_gemini.py
+++ b/src/langcheck/metrics/eval_clients/_gemini.py
@@ -105,13 +105,30 @@ class GeminiEvalClient(EvalClient):
                     "Please set the `GOOGLE_API_KEY` environment variable."
                 )
             self._client = genai.Client(vertexai=vertexai)
+            self._vertexai = vertexai
+
         else:
             self._client = genai_client
+            self._vertexai = genai_client.vertexai
+
+            # Client config will take precedence over the argument, and the
+            # argument will be ignored.
+            if self._vertexai and not vertexai:
+                warnings.warn(
+                    "Using an Vertex AI client but `vertexai` is False. "
+                    "Vertex AI client will be used."
+                )
+            elif not self._vertexai and vertexai:
+                warnings.warn(
+                    "Using Gemini Developer client but `vertexai` is True. "
+                    "Gemini Developer client will be used."
+                )
 
         if extractor is None:
             self._extractor = GeminiExtractor(
                 genai_client=self._client,
-                use_async=use_async,
+                use_async=self._use_async,
+                vertexai=self._vertexai,
             )
         else:
             self._extractor = extractor
@@ -282,6 +299,18 @@ class GeminiExtractor(Extractor):
             self._client = genai.Client(vertexai=vertexai)
         else:
             self._client = genai_client
+            # Client config will take precedence over the argument, and the
+            # argument will be ignored.
+            if genai_client.vertexai and not vertexai:
+                warnings.warn(
+                    "Using an Vertex AI client but `vertexai` is False. "
+                    "Vertex AI client will be used."
+                )
+            elif not genai_client.vertexai and vertexai:
+                warnings.warn(
+                    "Using Gemini Developer client but `vertexai` is True. "
+                    "Gemini Developer client will be used."
+                )
 
     def get_float_score(
         self,
--- a/src/langcheck/metrics/eval_clients/_gemini.py
+++ b/src/langcheck/metrics/eval_clients/_gemini.py
@@ -213,11 +213,7 @@ class GeminiSimilarityScorer(BaseSimilarityScorer):
                 )
                 return embed_response
 
-            try:
-                loop = asyncio.get_event_loop()
-            except RuntimeError:  # pragma: py-lt-310
-                loop = asyncio.new_event_loop()
-                asyncio.set_event_loop(loop)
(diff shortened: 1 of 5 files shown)


src/langcheck/metrics/eval_clients/_gemini.py
```diff
@@ -69,12 +76,42 @@ def __init__(
...
+                )
+            self._client = genai.Client(vertexai=vertexai)
+        else:
+            self._client = genai_client
 
         if extractor is None:
             self._extractor = GeminiExtractor(
+                genai_client=self._client,
                 use_async=use_async,
-                vertexai=vertexai,
```
yosukehigashi, line 77: nit: I guess it doesn't really matter since it will get ignored, but for consistency I think we should either (a) still pass this in anyway or (b) remove these from the Anthropic code.
```
        if extractor is None:
            self._extractor = AnthropicExtractor(
                anthropic_client=self._client,
                use_async=self._use_async,
                vertexai=self._vertexai,
            )
```

src/langcheck/metrics/eval_clients/_gemini.py
```diff
@@ -159,7 +196,11 @@ async def _call_async_api():
                 )
                 return embed_response
 
-            loop = asyncio.get_event_loop()
+            try:
+                loop = asyncio.get_event_loop()
+            except RuntimeError:  # pragma: py-lt-310
+                loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(loop)
```
yosukehigashi, line 203: (just for my understanding) what's the purpose of this part? does something not work on older python versions?

src/langcheck/metrics/eval_clients/_openai.py
```diff
@@ -48,20 +48,28 @@ def __init__(
...
+            if self._use_async and not use_async:
+                warnings.warn(
+                    "Using an AsyncOpenAI client but use_async is False. "
+                    "AsyncOpenAI client will be used."
+                )
+            elif not self._use_async and use_async:
+                warnings.warn(
+                    "Using OpenAI client but use_async is True. "
+                    "OpenAI client will be used."
+                )
```
yosukehigashi, line 61: these warnings are nice!! could we add these to the gemini and anthropic clients as well?

src/langcheck/metrics/eval_clients/_openai.py
```diff
@@ -48,20 +48,28 @@ def __init__(
...
         self._openai_args = openai_args
-        self._use_async = use_async
         self._system_prompt = system_prompt
 
         if extractor is None:
             self._extractor = OpenAIExtractor(
-                openai_client=self._client,
+                openai_client=openai_client,
                 openai_args=openai_args,
-                use_async=use_async,
```
yosukehigashi, line 64: same thing here - shouldn't we still pass in `use_async`? Otherwise if `openai_client` is async, the `OpenAIExtractor` constructor will output a warning

yosukehigashi (commented):
Nice work!! Just a few nits

src/langcheck/metrics/eval_clients/_gemini.py
yosukehigashi: after seeing the _openai.py changes, I'm leaning towards keeping these (i.e. option (a)). Especially if we add the warning logic to _anthropic.py and _gemini.py, we'll need to pass in the correct values to avoid hitting the warnings

src/langcheck/metrics/eval_clients/_gemini.py
taniokay: Oh sorry I was just testing something here, let me revert it.

src/langcheck/metrics/eval_clients/_gemini.py
taniokay: Agreed, I passed `vertexai` in Gemini client [1a79cb3](https://github.com/citadel-ai/langcheck/pull/199/commits/1a79cb380cfc081e2258c099b6eed21eb3ff4bd5) and `use_async` in OpenAI client [82fbaeb](https://github.com/citadel-ai/langcheck/pull/199/commits/82fbaebc86b02d34e2faaa91743dd42966d03848)!!

src/langcheck/metrics/eval_clients/_gemini.py
taniokay: BTW normally all we need to do is just call `asyncio.run(async_func())`, but `genai.Client fails` with that (calling `_embed` sequentially fails with `RuntimeError: Event loop is closed`). So as a workaround I'm getting `event_loop` explicitly

src/langcheck/metrics/eval_clients/_openai.py
taniokay: Added this to Gemini/Anthropic/OpenAI EvalClient & Extractor! [1a79cb3](https://github.com/citadel-ai/langcheck/pull/199/commits/1a79cb380cfc081e2258c099b6eed21eb3ff4bd5)

taniokay:
>  I pushed a change to the wording of the warning messages - please check that they make sense to you!

The warnings are much clearer now, thanks! I also bumped the dev version!

PR #201, opened 2025-05-27
Unpin transfomer upper version

liwii:
We pin the transformer version to `< 4.50.0` now. However, there is a vulnerability `transformers==4.49.0`, so we should unlock the newer version to avoid the security issues. This PR tries to resolve the local metric test errors caused by `transformers >= 4.50.0`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "langcheck"
-version = "0.10.0.dev3"
+version = "0.10.0.dev4"
 description = "Simple, Pythonic building blocks to evaluate LLM-based applications"
 readme = "README.md"
 authors = [{ name = "Citadel AI", email = "info@citadel.co.jp" }]
--- a/src/langcheck/__init__.py
+++ b/src/langcheck/__init__.py
@@ -1,4 +1,4 @@
 from langcheck import augment, metrics, plot, utils
 
 __all__ = ["augment", "metrics", "plot", "utils"]
-__version__ = "0.10.0.dev3"
+__version__ = "0.10.0.dev4"
--- a/src/langcheck/metrics/eval_clients/_anthropic.py
+++ b/src/langcheck/metrics/eval_clients/_anthropic.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import os
 import warnings
 from typing import Any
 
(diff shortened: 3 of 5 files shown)


liwii:
Updated the loading method of detoxify models so that it is compatible with `transformers >= 4.50.0`.
Ref: https://github.com/unitaryai/detoxify/blob/8f56f302bf8cf2673c2132fc2c2f5b2ca804815f/detoxify/detoxify.py

Confirmed that the outputs of en / ja detoxify models do not change on the following two examples used in pytest.

```
>>> from langcheck.metrics.en import toxicity
>>> from langcheck.metrics.ja import toxicity as ja_toxicity
>>> toxicity("I hate you. Shut your mouth!")
Progress: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 31.03it/s]
Metric: toxicity
              generated_outputs prompts sources reference_outputs explanations  metric_values
0  I hate you. Shut your mouth!    None    None              None         None        0.99031
>>> ja_toxicity("馬鹿")
Configuration Load Succeeded!
Progress: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 26.05it/s]
Metric: toxicity
  generated_outputs prompts sources reference_outputs explanations  metric_values
0                馬鹿    None    None              None         None       0.045536```

**Comment by liwii** [comment]
`pyright` complained about `fig.update_yaxes` or `fig.update_xaxes`, but as far as `plotly 6.0.1` documentation shows those attributes should be accessible. 

https://plotly.com/python/axes/
Also looked at CHANGELOG of > v6.0.1 but  no breaking changes show up.

Let me just `# type: ignore` for now.

**Comment by liwii** [comment]
The remaining error for factual consistency (ZH) model seems to be caused by the following change introduced in `transformers v4.52.3`. It is a warning about the model configuration, but the model doesn't even seem to follow the format this warning function expects.

https://github.com/huggingface/transformers/pull/37893

For now, we can pin the version to `tranformers < 4.52.3`. If the maintenance cost of this specific models become too high, we can also consider deprecating the model.

File: pyproject.toml [hunk 1]
```diff
@@ -30,7 +30,7 @@ dependencies = [
     'tomli; python_version < "3.11"',
     'tokenizers >= 0.13.2; python_version >= "3.11"',  # See https://github.com/citadel-ai/langcheck/pull/45
     'torch >= 2',
-    'transformers >= 4.6, < 4.50',
+    'transformers >= 4.6, < 4.52.1',
```
yosukehigashi, line 33: [comment] looks like `transformers.models` was introduced in 4.0 (https://github.com/huggingface/transformers/releases/tag/v4.0.0) so no need to update the lower bound

yosukehigashi (approved):
LGTM!

PR #202, opened 2025-05-29
Bump version to 0.10.0.dev5

liwii:
So that we can use the newer `tranformers` library to avoid the vulnerability issue

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "langcheck"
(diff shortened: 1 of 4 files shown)


yosukehigashi (approved):

yosukehigashi:
how did this fail but https://github.com/citadel-ai/langcheck/pull/201 pass..?

liwii:
?????

Let me take a look...

liwii:
Looks like they started to specify the torch dtypes as Literal rather than strings.
It doesn't cause that much practical problems, so let me `type: ignore` for now...

PR #208, opened 2025-07-19
Output reasoning summary

eycjur:
I added support for reasoning summaries from OpenAI and Azure OpenAI's reasoning models. However, reasoning summaries are only available via the response API, so I also made adjustments to use the response API when accessing reasoning summaries.

Resolves https://github.com/citadel-ai/citadel-lens/issues/4331

## Attention

1. Reasoning summary is only supported for certain models and API versions.
- Supported examples
  - openai: o1, o3-mini, o3, o4-mini, gpt-5
  - azure: o4-mioni, o3 (not deployed, so not verified), o3-mini(Not listed in the documentation but available), gpt-5
  - API Version: 2025-04-01-preview, preview
    - cf. https://learn.microsoft.com/en-us/azure/ai-foundry/openai/how-to/reasoning?tabs=python-secure%2Cpy#api--feature-support
2. Inference summaries are not generated for all requests.
- They tend not to be returned for simple problems, and this is particularly noticeable with advanced models such as o4-mini. (Fluency, toxicity, etc. are not returned at all. Multiplication of four digits is more likely to be returned.)        
  - cf. https://learn.microsoft.com/en-us/azure/ai-foundry/openai/how-to/reasoning?tabs=python-secure%2Cpy#reasoning-summary
3. Reasoning summary is in English.
- Even if the input is in Japanese, it seems to be returned in English.

## Verify operation
If you want to verify the operation, I recommend giving a calculation task that requires reasoning as a prompt as follows.

1. Create custom prompt template file (reasoning_summary.j2)
```j2:reasoning_summary.j2
Calculate 4325*3529 and answer whether the result is positive or negative.
```

2. Run evaluation script
```python
from langcheck.metrics.custom_text_quality import custom_evaluator
from langcheck.metrics.eval_clients import LiteLLMEvalClient

# Azure
eval_client = LiteLLMEvalClient(
    model="azure/<model>",
    api_key=...,
    api_base=...,
    api_version="2025-04-01-preview",
    use_reasoning_summary=True,
)

# OpenAI
eval_client = LiteLLMEvalClient(
    model="openai/<model>",
    api_key="sk-",  # your api key
    api_version="2025-04-01-preview",
    use_reasoning_summary=True,
) 

evaluator_score = custom_evaluator(
    generated_outputs=[""],
    prompts=None,
    sources=None,
    reference_outputs=None,
    eval_model=eval_client,
    metric_name="reasoning_summary",
    score_map=dict(positive=1, negative=0),
    template_path="reasoning_summary.j2",  # path to reasoning_summary.j2
    language="ja",
)
assert evaluator_score.explanations is not None
print(evaluator_score.explanations[0])
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "langcheck"
-version = "0.10.0.dev7"
+version = "0.10.0.dev8"
 description = "Simple, Pythonic building blocks to evaluate LLM-based applications"
 readme = "README.md"
 authors = [{ name = "Citadel AI", email = "info@citadel.co.jp" }]
--- a/src/langcheck/__init__.py
+++ b/src/langcheck/__init__.py
@@ -1,4 +1,4 @@
 from langcheck import augment, metrics, plot, utils
 
 __all__ = ["augment", "metrics", "plot", "utils"]
-__version__ = "0.10.0.dev7"
+__version__ = "0.10.0.dev8"


pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
     'litellm',
     'boto3',
     'nltk >= 3.9',
-    'openai >= 1',
+    'openai >= 1.92',
```
eycjur, line 27: This is unrelated to the current pull request, but I updated the version of the openai library as it seemed to have version restrictions.
The current `OpenAIExtractor` uses the parse method in `openai.resources.chat.completions.completions.Completions`. However, this is only supported in versions 1.92 and later.
In previous CI builds, versions 1.92 and later were installed, so the issue was not detected. However, for some reason, version 1.90 was installed this time, and the issue was discovered.

The CI that failed is here: https://github.com/citadel-ai/langcheck/actions/runs/16389481063/job/46432965886

eycjur:
@taniokay Can you review this pr?

eycjur:
@taniokay There is no strict deadline, so the priority is low, but I am reminding you because it may have been forgotten.

taniokay:
@eycjur Thanks for your friendly reminder, I'll take a look this weekend!

kennysong:
We'll need this for the GPT-5 models as well. They seem to refuse to explain their reasoning in the text output as well.

eycjur:
> We'll need this for the GPT-5 models as well. They seem to refuse to explain their reasoning in the text output as well.

This is an important point. However, we were able to generate inference summaries using both OpenAI and Azure's GPT-5 (no code changes were necessary). Note that Azure's GPT-5 seems to be unstable, and server errors occurred in some cases.

taniokay (commented):
Thanks for working on this, I confirmed that the eval client outputs now includes reasoning summary! 

I tried OpenAI models and it successfully outputs summary, but as for Azure models I got  `Resource not found` error with responses API, even though the it works with completion API. Do you know what's going wrong?

```
Failed to return an assessment corresponding to 0th prompt: litellm.APIError: AzureException APIError - {"error":{"code":"404","message": "Resource not found"}}
```

src/langcheck/metrics/eval_clients/_litellm.py
```diff
@@ -198,10 +246,40 @@ def get_text_responses(
...
+                            continue
+
+                        # Summary can be a list of summaries
+                        summaries.extend([s.text for s in output.summary])
+                    elif hasattr(output, "content"):
+                        content = output.content[0].text
+
+                if content is not None and summaries:
+                    summaries_str = "\n\n".join(summaries)
+                    content += f"\n\n**Reasoning Summary:**\n\n{summaries_str}"
```
eycjur, line 280: To return str, I add Reasoning Summary to content, but if there is a better way, we can fix it!

src/langcheck/metrics/eval_clients/_litellm.py
```diff
@@ -27,6 +28,10 @@ def __init__(
         embedding_model: str | None = None,
         *,
         use_async: bool = False,
+        use_reasoning_summary: bool = False,
+        reasoning_effort: ReasoningEffort = "medium",
+        reasoning_summary: Literal["auto", "concise", "detailed"]
+        | None = "auto",
```
taniokay, line 34: Related to [this comment](https://github.com/citadel-ai/langcheck/pull/208#discussion_r2266721125):

We can even remove `use_reasoning_summary` from the arguments since we can use `reasoning_summary` as a flag? In that case we need to make `reasoning_summary = None` by default

src/langcheck/metrics/eval_clients/_litellm.py
taniokay: I think this is clear enough to show in explanations, so let's keep it for now!

src/langcheck/metrics/eval_clients/_litellm.py
taniokay: Actually completions API also has `reasoning_effort` option, so let's keep `use_reasoning_summary` argument for clarity.

PR #210, opened 2025-09-08
allow template as string

anelephantbowl:


tests/metrics/test_custom_text_quality.py
```diff
@@ -62,6 +62,58 @@ def test_custom_evaluator(generated_outputs, sources, additional):
             assert metric_value == score_map[option]
 
 
+@pytest.mark.parametrize(
```
liwii, line 65: Can you add a test case to `custom_pairwise_evaluator` as well?

liwii (approved):
LGTM, just one nit

tests/metrics/test_custom_text_quality.py
anelephantbowl: [ec2bfe2](https://github.com/citadel-ai/langcheck/pull/210/commits/ec2bfe231e23d68c247b648fbc5a89fe0aed5551)

liwii:
Yeah I don't remember the details, but I think we intentionally disabled that kind of case for the custom metrics...

That doesn't align with the behavior of the default pairwise metrics, so removing the null check SGTM

Thanks for catching the issue with tempfile x unit test also!

PR #211, opened 2025-09-10
Bump the dev version

liwii:
To use #210 for internal development


liwii:
@anelephantbowl 

That's what we do before upgrading the internal package - make the "bumping dev version" commit and build a new wheel based on that

Can you take a quick look and approve it?

anelephantbowl (approved):
LGTM! Thanks for making this change!

PR #212, opened 2025-09-16
Remove synonym & install uv instead of devcontainer

taniokay:
This PR removes `synonym` augmentations and starts to use `uv` instead of dev container for environment setup.

`uv` finish the environment setup much faster, but Rust compiler was a blocker, which is required by synonym augmentations. I think it's okay to drop it considering its quality compared to the benefit from using uv


src/langcheck/augment/en/_synonym.py
```diff
@@ -1,56 +0,0 @@
-from __future__ import annotations
```
liwii, line 1: Do we need to remove `synonym` in the English module?

liwii (commented):
Sorry for the verrry late review! The developer experience with `uv` is much better, and I think removing `langcheck.ja.synonym` is fine.

`chikkarpy` is not maintained quite well, so it seems good to remove the dependency on that as well.

Left a comment about the English synonym.

Also, is it common to keep `uv.lock` in the repository?

PR #213, opened 2025-09-30
Handle None properly in LIteLLMExtractor.get_float_score

liwii:
Right now, many of the `get_float_score` implementations don't handle `None` in the input list properly: it actually sends `None` to the extractors and ask to map that to one of the options.

The expected behavior is to stop the call and return `None` instead. This PR applies this change to `LiteLLMEvalClient` (the other eval clients are not handled since we're planning to deprecate them anyways).

The behavior is confirmed by the newly added test case (it fails in the current `main`).

### DIFF
--- a/src/langcheck/metrics/eval_clients/_litellm.py
+++ b/src/langcheck/metrics/eval_clients/_litellm.py
@@ -493,7 +493,7 @@ class LiteLLMExtractor(Extractor):
             # A helper function to call the API with exception filter for alignment
             # of exception handling with the async version.
             def _call_api_with_exception_filter(
-                model_input: list[ChatCompletionMessageParam],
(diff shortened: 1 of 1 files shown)


liwii:
@taniokay @kennysong 
Can one of you guys take a look?

src/langcheck/metrics/eval_clients/_litellm.py
kennysong, line 493:
```suggestion
                model_input: list[ChatCompletionMessageParam] | None,
```

kennysong, line 493: model_input should probably be `list[ChatCompletionMessageParam] | None`

kennysong (approved):
LGTM, just one nit comment!

PR #214, opened 2025-10-06
Remove upper bound on transformers

kennysong:
Removed the upper bound on the `transformers` version since https://github.com/vllm-project/vllm-ascend/issues/2046 was resolved.

Also, the latest transformers dropped support for Python 3.9 (which reached EOL in October 2025), which was breaking the tests. Also updated the tests and minimum Python version to 3.10.


taniokay (approved):
LGTM!

